PR #4, opened 2020-07-22
frcnn_unit_test_code_refactor

sanjayuconn:
This commit includes following things:
- frcnn code refactor
- frcnn unit test cases
- frcnn doc string


datasetinsights/data/dummy/dummy_object_detection.py
```diff
@@ -0,0 +1,40 @@
+
+import numpy as np
+from PIL import Image
+
+from datasetinsights.data.datasets import Dataset
+from datasetinsights.data.bbox import BBox2D
+
+
+class DummyDetection2D(Dataset):
```
adason, line 10: We should have docstrings for this dataset.

datasetinsights/data/dummy/dummy_object_detection.py
```diff
@@ -0,0 +1,40 @@
...
+            Image.fromarray(np.random.random(image_size), "L")
+        ]
+        self.bboxes = [
+            [BBox2D(label=1, x=10, y=20, w=30, h=40),
+             BBox2D(label=2, x=50, y=50, w=10, h=10)],
+            [BBox2D(label=1, x=30, y=40, w=20, h=20),
+             BBox2D(label=2, x=20, y=10, w=40, h=10)]
+        ]
+        self.label_mappings = label_mappings
+        if not self.label_mappings:
```
adason: Why not just leave this as the default input value? Since we don't provide users with a lot of control over this dummy dataset, should `label_mappings` be input arguments for `DummyDetection2D ` class?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,379 @@
+import torch
```
adason: Should run isort here as per contribution [guide](https://github.com/Unity-Technologies/dataset-insights/blob/master/CONTRIBUTING.md#style-guide).

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
 logger = logging.getLogger(__name__)
 DEFAULT_ACCUMULATION_STEPS = 1
+TRAIN = "train"
+VAL = "val"
+TEST = "test"
+
+
+class FasterRCNN(Estimator):
+    """
+    https://github.com/pytorch/vision/tree/master/references/detection
```
adason, line 36: This docstring should be improved. We should describe what the estimator is (instead of just a few links to model without some user friendly descriptions.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+    https://arxiv.org/abs/1506.01497
+    Args:
+        config (CfgNode): estimator config
+        writer: Tensorboard writer object
+        checkpointer: Model checkpointer callback to save models
+        device: model training on device (cpu|cuda)
```
adason, line 43: Args doc mismatch with what's in the `__init__` method. Can you improve it?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+        local_rank (int): (optional) rank of process executing code
+        gpu (int): (optional) gpu id on which code will execute
+    Attributes:
```
adason, line 47: Attributes doc mismatch with public instance variables for this class. Can you improve it?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+            )
+        self.model.to(self.device)
+
+        if self.config.system.distributed:
+            self.model = torch.nn.parallel.DistributedDataParallel(
+                self.model, device_ids=[self.gpu]
+            )
+            self.model_without_ddp = self.model.module
+
+    def train(self, **kwargs):
```
adason, line 106: Docstring? Whant's the input parameter here?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+        val_dataset = download_data(config, VAL)
+        label_mappings = train_dataset.label_mappings
+        if self.config.system.dryrun:
+            train_dataset = create_dryrun_dataset(config, train_dataset, TRAIN)
+            val_dataset = create_dryrun_dataset(config, val_dataset, VAL)
+
+        logger.info(f"length of train dataset is {len(train_dataset)}")
+        logger.info(f"length of validation dataset is {len(val_dataset)}")
+        is_distributed = config.system.distributed
+        train_sampler = create_sampler(is_distributed=is_distributed,
```
adason: Why is `create_sampler` a module level method? Can this be staticmethod for the FasterRCNN class?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+                                       dataset=train_dataset,
+                                       is_train=True)
+        val_sampler = create_sampler(is_distributed=is_distributed,
+                                     dataset=val_dataset,
+                                     is_train=False)
+
+        train_loader = create_dataloader(config, train_dataset, train_sampler,
```
adason: Why is create_sampler a module level method? Can this be staticmethod for the FasterRCNN class?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+    def train_loop(
+        self,
+        *,
+        train_dataloader,
+        label_mappings,
+        val_dataloader,
+        train_sampler=None,
+    ):
+        """
+        Args:
```
adason, line 147: The docstring should be improved here if this is a public method. Alternatively, you can make this private method and be more relaxed on the docstring.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,594 @@
...
+            self,
+            *,
+            optimizer,
+            data_loader,
+            epoch,
+            lr_scheduler,
+            accumulation_steps,
+    ):
+        """
+
```
adason, line 204: Improve docstring here. It needs to have title and description about this public method.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+        self,
+        *,
+        data_loader,
+        epoch,
+        label_mappings,
+        max_detections_per_img=MAX_BOXES_PER_IMAGE,
+        is_distributed=False,
+        synchronize_metrics=True,
+    ):
+        """
```
adason: Please follow [pep-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) docstring standard.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+        rcnn requires input and gt data for training mode and returns a
+        dictionary of losses (which we need to record the loss). We also need to
+        get the raw predictions, which is only possible in model.eval() mode,
+        to calculate the evaluation metric.
+        Args:
+            data_loader (DataLoader): pytorch dataloader
+            epoch (int): current epoch, used for logging
+            label_mappings (dict): a dict of {label_id: label_name} mapping
+            max_detections_per_img: max number of targets or predictions allowed
+            per example
```
adason, line 314: [nit] might be better to indent this line to improve readability.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+    Returns dataset: Subset dataset
+
+    """
+    r = np.random.default_rng()
+    indices = r.integers(0, len(dataset), config[split].batch_size * 2)
+    dataset = torch.utils.data.Subset(dataset, indices)
+    return dataset
+
+
+def download_data(config, split):
```
adason: Why do we need a 1-line method to download dataset? This method seems to be over-engineering.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+    Returns:
+        data_sampler: (torch.utils.data.Sampler)
+
+    """
+    if is_distributed:
+        sampler = torch.utils.data.distributed.DistributedSampler(
+            dataset, shuffle=is_train
+        )
+
+    else:
```
adason: There might be some opportunity to flatten this nested if statements. Would this improve code readability?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+        if is_train:
+            sampler = torch.utils.data.RandomSampler(dataset)
+        else:
+            sampler = torch.utils.data.SequentialSampler(dataset)
+    return sampler
+
+
+def load_data(
```
adason: How is `load_data` different from `create_dataloader`? Should we rename both methods to improve readability.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+        dataset=dataset,
+        batch_size=config[split].batch_size,
+        sampler=sampler,
+        collate_fn=collate_fn,
+        train=is_train,
+    )
+    return dataloader
+
+
+def create_optimizer_lrs(config, model):
```
adason: Should this method be a module level method? Should it be static method for the FasterRCNN class?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+    """
+    creates optimizer and learning rate scheduler
```
adason: PEP-257

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -22,6 +22,605 @@
...
+    Args:
+        config: (CfgNode): estimator config:
+        model: pytorch model
```
adason: Might be simpler pass `parameters` instead of the whole pytorch model?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -110,12 +709,13 @@ class label for each ground-truth box. If the dictionary has the key
 
 def reduce_dict(input_dict, average=True):
     """
+    Reduce the values in the dictionary from all processes so that all processes
```
adason, line 751: PEP-257

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -164,84 +764,13 @@ def prepare_bboxes(bboxes: List[BBox2D]) -> Dict[str, torch.Tensor]:
...
-
-
 def canonical2list(bbox: BBox2D):
     """
     convert a BBox2d into a single list
     Args:
         bbox:
 
-    Returns:
+    Returns: attribute list of BBox2D
```
adason: Please follow google style docstring as per contribution guide.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
-        """
-
-        Args:
-            dataset: dataset obj must have len and __get_item__
-            is_train: whether or not the sampler is for training data
 
-        Returns:
-            https://pytorch.org/docs/stable/data.html#torch.utils.data.Sampler
+def gather_gt_preds(*, gt_preds, device, max_boxes=MAX_BOXES_PER_IMAGE):
+    """
```
adason: PEP-257

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
-        logger.info(f"length of train dataset is {len(train_dataset)}")
-        val_dataset = Dataset.create(
-            config.val.dataset.name,
-            data_root=config.system.data_root,
-            transforms=get_transform(),
-            **config.val.dataset.args,
-        )
-        logger.info(f"length of validation dataset is {len(val_dataset)}")
+def tensorlist2canonical(tensor_list):
+    """
```
adason: PEP-257

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
+    return gt_preds_canonical
 
-    def save(self, path):
-        """ Serialize Estimator to path
 
-        Args:
-            path (str): full path to save serialized estimator
+def metric_per_class_plot(metric_name, data, label_mappings, figsize=(20, 10)):
+    """
+    Bar plot for metric per class.
```
adason: PEP-257

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
 
-        Returns:
-            saved full path of the serialized estimator
-        """
-        save_dict = {
-            "model": self.model_without_ddp.state_dict(),
-            "config": self.config,
-        }
-        torch.save(save_dict, path)
+    Returns (matplotlib.pyplot.figure): a bar plot for metric per class.
```
adason: Follow google style docstring

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
+    )
+    plt.margins(0.2)
+    plt.subplots_adjust(bottom=0.3)
+    return fig
 
-        return path
 
-    def load(self, path):
-        """ Load Estimator from path
+class PrepBoxes:
```
adason: This seems to be an opportunity combining `ToTensor` and `PrepBoxes`. One is responsible of converting PIL image to tensor the other is responsible of converting BBox2D objects to tensor.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
-        """
-        pil_img = pil_img.unsqueeze(0)
-        img_tensor = pil_img.to(self.device)
-        predicts = self.model_without_ddp(img_tensor)
-        predict_annotations = convert_bboxes2canonical(predicts)[0]
-        filtered_pred_annotations = [
-            box for box in predict_annotations if box.score >= box_score_thresh
-        ]
-        return filtered_pred_annotations
+def get_transform():
```
adason: Why is this a module level method? Maybe it can be a staticmethod of FasterRCNN class?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -355,619 +884,123 @@ def _gt_preds2tensor(gt_preds, max_boxes=MAX_BOXES_PER_IMAGE):
...
-                optimizer,
-                step_size=config.optimizer.args.lr_step_size,
-                gamma=config.optimizer.args.lr_gamma,
-            )
-        else:
-            raise ValueError(
-                f"only valid optimizers are 'SGD' and 'Adam' but "
-                f"received paramater {config.optimizer.name}"
-            )
+def collate_fn(batch):
```
adason: Is this method required as it's only a 1-line method?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+    create_dryrun_dataset,
+    create_optimizer_lrs,
+    create_sampler,
+    download_data,
+    get_transform,
+    load_data,
+)
+from datasetinsights.storage.checkpoint import create_checkpointer
+
+TRAIN = "train"
```
adason: Why do you need to redefine these constants? Can you read from estimators/faster_rcnn.py?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+VAL = "val"
+TEST = "test"
+
+label_mappings = {"1": "car", "2": "bike"}
```
adason: Why do you need these global variables in tests? Shouldn't they be independent for each unittests?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+dummy_data = DummyDetection2D(
+    transform=get_transform(), label_mappings=label_mappings
+)
+
+
+@pytest.fixture
+def config():
+    parser = argparse.ArgumentParser(
```
adason: Why do you need `ArgumentParser` here? This test is not going to run with CLI arguments? Can this method simply return  `cfg`?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+    load_data,
+)
+from datasetinsights.storage.checkpoint import create_checkpointer
+
+TRAIN = "train"
+VAL = "val"
+TEST = "test"
+
+label_mappings = {"1": "car", "2": "bike"}
+dummy_data = DummyDetection2D(
```
adason: Should probably make this a pytest fixture.

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+    )
+    cfg = CN.load_cfg(open(cmd_args.config, "r"))
+
+    return cfg
+
+
+@patch("datasetinsights.estimators.faster_rcnn.Dataset.create")
+def test_faster_rcnn_train_one_epoch(mock_create, config):
+    mock_create.return_value = dummy_data
+    config.system.logdir = "/tmp/train_one/"
```
adason: This path is os dependent. I will suggest using `tempdir.TemporartDirectory()`. This method will give you different temporary directory everytime you run a single unit of unittest.

datasetinsights/data/dummy/dummy_object_detection.py
sanjayuconn: Fixed

tests/test_faster_rcnn.py
sanjayuconn: Did all three linting check before PR, still it has. 
black .
isort -rc .
flake8

datasetinsights/data/dummy/dummy_object_detection.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Fixed

tests/test_faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: [nit]?

adason: Just mean this is nit-picking comment in the context of PR

https://stackoverflow.com/questions/27810522/what-does-nit-mean-in-hacker-speak

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+    accumulation_steps = config.train.get("accumulation_steps", 1)
+    estimator.train_one_epoch(
+        optimizer=optimizer,
+        data_loader=train_loader,
+        epoch=0,
+        lr_scheduler=lr_scheduler,
+        accumulation_steps=accumulation_steps,
+    )
+
+    assert os.listdir("/tmp/train_one/")[0].startswith("events.out.tfevents")
```
adason: Rely on output files that was specific for a training procedure is a test that need to be changed if the output format is somehow changed (for example,tensorboard api change). A better test usually requires validate the code execution with the mocked objects. 

For example, in this test, we can improve this test with some mocks:

```
def test_faster_rcnn_train_one_epoch(config):
    writer = Mock()
    kfp_writer = Mock()
    checkpointer = Mock()
    device = torch.device("cpu")

    estimator = FasterRCNN(
        config=config,
        writer=writer,
        device=device,
        checkpointer=checkpointer,
        kfp_writer=kfp_writer,
    )

    dataset = DummyDetection2D()
    sampler = torch.utils.data.RandomSampler(dataset)
    data_loader = torch.utils.data.DataLoader(dataset, sampler)

    # ideally this should be mocked as well...
    optimizer, lr_scheduler = create_optimizer_lrs(config, estimator.model)
    accumulation_steps = config.train.get("accumulation_steps", 1)
    epoch = 0
    estimator.train_one_epoch(
        optimizer=optimizer,
        data_loader=data_loader,
        epoch=epoch,
        lr_scheduler=lr_scheduler,
        accumulation_steps=accumulation_steps,
    )
    loss_value = 0.1
    with @patch("datasetinsights.estimator.faster_rcnn.Loss.compute") as mock_loss_compute:
        mock_loss_compute.return_value = loss_value
        writer.assert_called_with("training/loss", loss_value, epoch)
        writer.assert_called_with("training/running_loss", loss_value, epoch)

```

This is not a perfect solution either, as it might need to be fixed when the training code is changed. However, at least we know the test will not fail when external packages are changed. 

Does this make sense? Does this thinking apply to other test cases?

tests/test_faster_rcnn.py
```diff
@@ -0,0 +1,377 @@
...
+        config.estimator,
+        config=config,
+        writer=writer,
+        device=torch.device("cpu"),
+        checkpointer=checkpointer,
+        kfp_writer=kfp_writer,
+    )
+    optimizer, lr_scheduler = create_optimizer_lrs(config, estimator.model)
+
+    assert optimizer
```
adason: What's the purpose of this assert? It will always be true if the object exists. I think a better test that more robust to external API change can be:

```
def test_optimizer_lr():
    # mock torch.optim.Adam
    # mock torch.optim.lr_scheduler.LambdaLR
    config = {...name: Adam, args: {...}}
    op, lr = create_optimizer_lrs(config, params)
    
    assert mock_adam == op
    assert mock_lr = lr
    op.assert_called_with(...)
```

Does this pattern applies to other static methods or module level methods?

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: No input parameter, its simple method call on object.

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Yes

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: yes

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: It seems one line, but it has alot parameters so makes bit lengthy main method and naming(Dataset.create) makes bit unreadable so fix both created separate method.

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: yes, fixed!

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: create_dataloader triggers the data load process, load_data loads the data. We can put in same method but it will bit lengthy. Spliting into two units make more readable I think.

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: why to combine both? both have different purpose? and it has call implementation, which does the object content transformation.

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: This we can keep outside, this will be called from many places where data transformations requires like dummy data loader and others.

datasetinsights/estimators/faster_rcnn.py
sanjayuconn: yeah its one line, but I think we can keep. it does transformation of dynamic batches and been used many places so for readability we can keep.

tests/test_faster_rcnn.py
sanjayuconn: Fixed

tests/test_faster_rcnn.py
sanjayuconn: How to initialize parser then?

tests/test_faster_rcnn.py
sanjayuconn: Fixed

datasetinsights/estimators/faster_rcnn.py
BlairLee: Better to rename both methods to be more readable.

tests/test_faster_rcnn.py
sanjayuconn: I tried this, approach, by mocking both writer and checkpointer but below assertions didn't get called.    

writer.assert_called_with("training/loss", loss_value, epoch)
 writer.assert_called_with("training/running_loss", loss_value, epoch).

tests/test_faster_rcnn.py
BlairLee: You can use [yacs](https://github.com/rbgirshick/yacs) to do the initialization. And you can put `config` in a `conftest.py` ([example](https://www.tutorialspoint.com/pytest/pytest_conftest_py.htm)).

tests/configs/faster_rcnn_groceries_real_test.yaml
```diff
@@ -0,0 +1,60 @@
...
+  data_root: /tmp
+  distributed: False
+  dryrun: True
+  gpus_per_node: 1
+  logdir: /tmp
+  no_cuda: False
+  val_interval: 1
+  verbose: False
+  workers: 0
+
```
BlairLee: Why do we need a separate yaml file? Can we use the original one and mock `system` args separately?

datasetinsights/data/dummy/dummy_object_detection.py
```diff
@@ -0,0 +1,53 @@
...
+from datasetinsights.data.datasets import Dataset
+
+
+class DummyDetection2D(Dataset):
+    """Dummy dataset creation with 2 images for testing.
+
+    Args:
+        image_size : size of images you want generate
+        transform : transform the images and bounding boxes
+
```
adason, line 18: What are the attributes in this class? Can you improve the docstring?

tests/test_faster_rcnn.py
adason: This has to be fixed. It's absolutely not required to have argparse stuff in unittests, as there are no way for these arguments to be passed during the test execution.

datasetinsights/estimators/faster_rcnn.py
adason: Can the method name be improved? It's very ambiguous to see a name like "download_data" as this not what this method does.

datasetinsights/estimators/faster_rcnn.py
adason: Agree with @BlairLee

datasetinsights/estimators/faster_rcnn.py
adason: Maybe this calls for renaming PrepBoxes? Can it be more explicit? I am OK if we have steps that handle bounding boxes in a format separate from ToTensor code, but then I will argue why is `ToTensor` method not responsible for converting from bounding box is a certain format to pytorch tensor? With this design, maybe we can simple replace `ToTensor` with `ToTFTensor` so that we can support this method for tensorflow?

datasetinsights/estimators/faster_rcnn.py
adason: The question to answer is, is this method specific for FasterRCNN? If the answer is yes, why is it in the module? Is there other types of model that will use these transformations?

datasetinsights/estimators/faster_rcnn.py
adason: So should this method kept in faster_rcnn module? Maybe it does not belong here?

tests/test_faster_rcnn_util.py
```diff
@@ -272,3 +293,199 @@ def test_convert2canonical_batch():
...
+
+    actual_result = reduce_dict(input_dict)
+    expected_result = {
+        "loss_classifier": 0.1423666626214981,
+        "loss_box_reg": 0.1423666626214981,
+        "loss_objectness": 0.1423666626214981,
+        "loss_rpn_box_reg": 0.1423666626214981,
+    }
+
+    assert all(
```
adason, line 333: [nit] why not use `assert actual_result  == expected_result `

tests/test_faster_rcnn_util.py
```diff
@@ -272,3 +293,199 @@ def test_convert2canonical_batch():
...
+    expected_result = [0, 1.0, 10, 10, 10, 10]
+    assert actual_result == expected_result
+
+
+def test_list2canonical():
+    """test list to canonical."""
+    input_list = [0, 1.0, 10, 10, 10, 10]
+    actual_result = list2canonical(input_list)
+
+    assert all(
```
adason, line 362: [nit] why not use [asserlistequal](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertListEqual) or simply` assert input_list == actual_result`?

adason (approved):

PR #5, opened 2020-07-22
Make GrocereisReal dataset public

BlairLee:
move GrocereisReal to public path and add checksums.


datasetinsights/data/simulation/download.py
```diff
@@ -289,6 +296,58 @@ def _download_file(source_uri: str, dest_path: str, use_cache: bool = True):
     return dest_path
 
 
+def compare_checksums(file_path, checksum_path):
```
adason: I will suggest adding unttests for `compare_checksum` method.

adason (changes requested):

datasetinsights/scripts/public_download.py
```diff
@@ -0,0 +1,89 @@
...
+    parser.add_argument(
+        "--version", type=str, default="v3", help="dataset version, e.g. v3",
+    )
+
+    args = parser.parse_args()
+
+    return args
+
+
+def run(args):
```
adason: Can you add tests for this method? It might be easier to refactor this method so that it's easier to test.

BlairLee: Since all the methods in `run()` have been tested and all the parameters are constant, I don't think it needs a unittest.

adason: We should mock up code that was already tested. However, there are still values to test how each of these methods are wired up together.

tests/simulation/test_download.py
```diff
@@ -154,3 +156,16 @@ def test_match_filetypes():
...
     assert Downloader.match_filetypes(manifest) == expected_filetypes
+
+
+def test_compare_checksums():
+    parent_dir = pathlib.Path(__file__).parent.parent.absolute()
+    file_path = str(parent_dir / "mock_data" / "calib000000.txt")
+    checksum_path = str(parent_dir / "mock_data" / "checksum_calib000000.txt")
+
+    raise_error = False
+    try:
```
adason: Should use pytest to assert raises like [this](https://docs.pytest.org/en/3.0.1/assert.html#assertions-about-expected-exceptions)

datasetinsights/scripts/public_download.py
BlairLee: Done.

tests/simulation/test_download.py
BlairLee: Done.

adason: Why do you need a `no_raises` method? If the checksum passes, it should return True/False to validate it passed. From the method called `compare_checksums` it implies the method will return the checksum results (whether it passes/failed)

tests/simulation/test_download.py
```diff
@@ -154,3 +159,44 @@ def test_match_filetypes():
...
+    parser.add_argument("--data-root", type=str, default="")
+    parser.add_argument(
+        "--name",
+        type=str,
+        default="GroceriesReal",
+        required=True,
+    )
+    parser.add_argument("--version", type=str, default="v3")
+
+    args = parser.parse_args()
```
adason: Why do you need to use `ArgumentParser` here? Can you directly mock up an `args` class here? 
If in the future, we added more public dataset to this method, does that mean we have to completely changer this test? Is there a better way to make it open for extension?

adason (changes requested):

tests/simulation/test_download.py
BlairLee: But if we just return True/False, every time we do the checksum, we need to raise error:
```
if not compare_checksums(
        file_path=file_dest_path, checksum_path=checksum_dest_path
    ):
        raise DownloadError(
            f"Invalid hash value."
            f"There are some errors during the dataset download. "
            f"Please download it again."
        )
```

tests/simulation/test_download.py
BlairLee: Use `CN()` to mock args.

datasetinsights/data/datasets/exceptions.py
```diff
@@ -0,0 +1,3 @@
+class DatasetError(Exception):
```
adason: [nit] Might be better renaming this to `DatasetNotFoundError` or `DatasetMissingError`?

datasetinsights/data/datasets/groceries_real.py
```diff
@@ -157,6 +176,86 @@ def _filepath(self, filename):
...
+    @staticmethod
+    def _download_http(source_uri, dest_path, version):
+        """ Download dataset from Public HTTP URL.
+
+        Args:
+            source_uri (str): source url where the file should be downloaded
+            dest_path (str): destination path of the file
+
+        Raises:
+            DownloadError if a given definition path can't be found
```
adason: This description is not accurate. When will `DownloadError` be raised according the the login implemented in this method?

datasetinsights/data/datasets/groceries_real.py
```diff
@@ -157,6 +176,86 @@ def _filepath(self, filename):
...
+            zip_dir.extractall(root_dir)
+
+    @staticmethod
+    def download(data_root, version):
+        """ Download dataset from Public HTTP URL.
+
+        If the file already exists and the checksum matches, it will skip the
+        download step. If not, it would delete the previous file and download
+        it again. If the file doesn't exist, it would download the file.
+        """
```
adason, line 235: This docstring is incomplete. What's input args and what's the retrun type? What's the expected error that might be raised?

datasetinsights/data/datasets/groceries_real.py
```diff
@@ -157,6 +176,86 @@ def _filepath(self, filename):
...
+        valid_versions = tuple(
```
adason: This conversion is not required. You should be able to check if a give key exists in a python dictionary.

datasetinsights/data/download.py
```diff
@@ -64,3 +65,47 @@ def download_file(source_uri: str, dest_path: str, use_cache: bool = True):
...
+    """ Compute the checksum of a file.
+
+    Args:
+        filepath (str): the doaloaded file path
+        algorithm (str): checksum algorithm. Defaults to CRC32
+
+    Returns:
+        int: the checksum value
+    """
+    chs = -1
```
adason: Why do we need a default checksum value `-1`? If the checksum computation failed, doesn't it simply raise some error?

datasetinsights/data/datasets/groceries_real.py
BlairLee: It will be raised when there are HTTP requests' exceptions. Will change the description.

datasetinsights/data/download.py
BlairLee: Yes, it does. Will remove the initialization in case there is a checksum of a file equals -1.

datasetinsights/data/datasets/groceries_real.py
```diff
@@ -157,6 +176,92 @@ def _filepath(self, filename):
...
+        expected_checksum = GroceriesReal.GROCERIES_REAL_DATASET_TABLES[
+            version
+        ].checksum
+        extract_folder = os.path.join(data_root, GroceriesReal.LOCAL_PATH)
+        if os.path.exists(dest_path):
+            logger.info("The dataset file exists. Skip download.")
+            try:
+                validate_checksum(dest_path, expected_checksum)
+            except ChecksumError:
+                logger.info(
```
adason, line 253: [nit] Since this hides ChecksumError from execution, maybe it's also worth to log error message as warning using `logger.warning(ex)`

adason (approved):

PR #6, opened 2020-07-24
Download Method Refactor

Saurav-D:
# Peer Review Information

 - Moved `_download_file` method from `datasetinsights.data.simulations.download` to `datasetinsights.data.download`
 - Made `_download_file method` public (now `download_file`)
 - Moved `DownloadError` from  `datasetinsights.data.simulations.exceptions` to `datasetinsights.data.exceptions`

# Pull Request Check List

- [x] Updated **tests** for changed code.

### DIFF
--- a/datasetinsights/data/download.py
+++ b/datasetinsights/data/download.py
@@ -1,0 +1,66 @@
+import logging
+from pathlib import Path
+
+import requests
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+from .exceptions import DownloadError
+
+logger = logging.getLogger(__name__)
+
+# Timeout of requests (in seconds)
+DEFAULT_TIMEOUT = 1800
+# Retry after failed request
+DEFAULT_MAX_RETRIES = 5
+
+
+class TimeoutHTTPAdapter(HTTPAdapter):
+    def __init__(self, timeout, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


datasetinsights/data/download.py
```diff
@@ -0,0 +1,66 @@
...
+                f"be completed."
+            )
+
+            raise DownloadError(err_msg)
+        else:
+            dest_path.parent.mkdir(parents=True, exist_ok=True)
+            with open(dest_path, "wb") as f:
+                f.write(response.content)
+
+    return dest_path
```
BlairLee, line 66: Can we rename this file name, like `http_download.py`? It's wired to have the same name with `datasetinsights/data/simulation/download.py`.

Saurav-D: yes, we can do that. But this is contains generic download function on the other hand, `datasetinsights/data/simulation/download.py` contains download functions related to USIM so we can also change that to `usim_download.py`

BlairLee: Ok. Sounds good!

Saurav-D: I talked with YC about this, and we think it's fine to have `download.py` in both places.

BlairLee: Okay. I'm okay with that.

BlairLee (approved):

PR #8, opened 2020-07-27
Refactored EstimatorCheckpoint

Saurav-D:
# Peer Review Information

- Refactored EstimatorCheckpoint Class
- Added Writer classes for GCS and Local Estimator Checkpoint
- Added Loader methods for GCS, Local and HTTP Estimator Checkpoint

# Pull Request Check List

- [ ] Added and updated **tests** for changed code.

### DIFF
--- a/datasetinsights/storage/checkpoint.py
+++ b/datasetinsights/storage/checkpoint.py
@@ -16,17 +16,17 @@ DEFAULT_SUFFIX = "estimator"
 
 
 class EstimatorCheckpoint:
-    """
-    For loading and saving estimator checkpoints
+    """For loading and saving estimator checkpoints
 
     Args:
         estimator_name (str): name of the estimator
         log_dir (str): log directory
         distributed (bool): boolean to determine distributed training
+
     Attributes:
-        estimator_name (str): name of the estimator
         log_dir (str): log directory
         distributed (bool): boolean to determine distributed training
+
     """
 
     def __init__(self, estimator_name, log_dir, distributed):
@@ -38,7 +38,7 @@ class EstimatorCheckpoint:
         if log_dir.startswith(const.GCS_BASE_STR):
             writer = GCSEstimatorWriter(log_dir, estimator_name)
         else:
-            writer = LocalEstimatorWriter(estimator_name)
+            writer = LocalEstimatorWriter(log_dir, estimator_name)
 
         return writer
 
@@ -52,27 +52,28 @@ class EstimatorCheckpoint:
             method = load_local
         else:
             raise ValueError(
-                f"Given path: {0}, is either invalid or not "
-                "supported".format(path)
+                f"Given path: {path}, is either invalid or not supported."
+                f"Currently supported path are local path (/path/to/file), "
+                f"GCS path (gs://) and HTTP url (http:// or https://) path"
             )
 
         return method
 
     def save(self, estimator, epoch):
-        """
-        Saves estimator to log directory
+        """Saves estimator to log directory
 
         Args:
             estimator (datasetinsights.estimators.Estimator):
             datasetinsights estimator object
             epoch (int): epoch number
+
         """
         if self.distributed and not is_master():
             return
         self._writer.save(estimator=estimator, epoch=epoch)
 
     def load(self, estimator, path):
-        """ Loads estimator from given path
+        """Loads estimator from given path
 
         Path can be either a local path or GCS path or HTTP url
 
@@ -80,6 +81,7 @@ class EstimatorCheckpoint:
             estimator (datasetinsights.estimators.Estimator):
             datasetinsights estimator object
             path (str): path of estimator
+
         """
         if self.distributed and not is_master():
             return
@@ -89,7 +91,7 @@ class EstimatorCheckpoint:
 
 
 class LocalEstimatorWriter:
-    """ Writes (saves) estimator checkpoints locally
+    """Writes (saves) estimator checkpoints locally
 
     Args:
         prefix (str): filename prefix of the checkpoint files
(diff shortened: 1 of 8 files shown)


datasetinsights/storage/checkpoint.py
```diff
@@ -5,89 +5,109 @@
...
 DEFAULT_SUFFIX = "estimator"
-GCS_BASE_STR = "gs://"
 
 
-def checkpoint_file_on_gcs(config):
+class EstimatorCheckpoint:
     """
-    determine whether or not the checkpoint file to load an estimator is on
-    gcs
+    For loading and saving estimator checkpoints
```
adason: Please follow milti-line docstring in [pep-257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). Also, since this is the main interface that users will interact with, it deserves better docstrings describing different behaviors of this class.

datasetinsights/storage/checkpoint.py
```diff
@@ -5,89 +5,109 @@
...
 
     Args:
-        config:
+        estimator_name (str): name of the estimator
+        log_dir (str): log directory
+        distributed (bool): boolean to determine distributed training
+    Attributes:
```
adason, line 31: Please update Attributes to reflect recent changes.

datasetinsights/storage/checkpoint.py
```diff
@@ -5,89 +5,109 @@
...
+    def _get_loader_from_path(path):
+        if path.startswith((const.HTTP_URL_BASE_STR, const.HTTPS_URL_BASE_STR)):
+            method = load_from_http
+        elif path.startswith(const.GCS_BASE_STR):
+            method = load_from_gcs
+        elif path.startswith("/"):
+            method = load_local
+        else:
+            raise ValueError(
+                f"Given path: {0}, is either invalid or not "
```
adason: Use python's f-string and remove instead of `"".format(...)`. Also, this error should be more informative to tell users what. type of paths are supported.

datasetinsights/storage/checkpoint.py
```diff
@@ -5,89 +5,109 @@
...
 
-def create_checkpointer(*, logdir, config):
-    """
-    Initialize the correct estimator checkpointer
-    Args:
-        logdir: filepath to where to save/load local copy of estimator
-        config:
+    def save(self, estimator, epoch):
+        """
+        Saves estimator to log directory
```
adason: PEP-257

datasetinsights/storage/checkpoint.py
```diff
@@ -5,89 +5,109 @@
...
         prefix (str): filename prefix of the checkpoint files
         suffix (str): filename suffix of the checkpoint files
     """
 
-    def __init__(
-        self, dirname, prefix, *, suffix=DEFAULT_SUFFIX, create_dir=True
-    ):
-        self.dirname = dirname
+    def __init__(self, prefix, *, suffix=DEFAULT_SUFFIX):
+        self.dirname = os.path.join(const.PROJECT_ROOT, prefix)
```
adason: Why do we have to remove the original and assumes estimator will create under `PROJECT_ROOT`? The original design allow user to use any estimator dir to any location.

datasetinsights/storage/checkpoint.py
```diff
@@ -162,32 +169,47 @@ def save(self, estimator, epoch=None):
...
-        if self.is_master:
-            self.estimator_checkpoint.save(estimator, epoch)
+    Args:
+        estimator (datasetinsights.estimators.Estimator):
+            datasetinsights estimator object.
+        full_cloud_path: full path to the checkpoint file
+    """
+    bucket, object_key = gcs_bucket_and_path(full_cloud_path)
+    filename = os.path.basename(object_key)
+    temp_dir = tempfile.TemporaryDirectory().name
```
adason: How do you make sure to cleanup `TemporaryDirectory` after model artifact is loaded into memory? Using context manager might help. See these [examples](https://docs.python.org/3/library/tempfile.html#tempfile-examples).

tests/test_checkpoint.py
```diff
@@ -17,47 +18,34 @@ def create_empty_file(path):
     path.touch()
 
 
-def test_estimator_checkpoint_creates_dir():
+def test_local_estimator_writer_creates_dir():
     prefix = "good_model"
-    with tempfile.TemporaryDirectory() as tmp_dir:
-        dirname = os.path.join(tmp_dir, "some_dir_name")
-        EstimatorCheckpoint(dirname, prefix)
+    dirname = os.path.join(const.PROJECT_ROOT, prefix)
```
adason: It might be better teses cleanup itself after runs. I will recommdend deleting these directory after test finished. Using `TemporaryDirectory()` helps to cleanup directory after each unittest runs.

tests/test_checkpoint.py
```diff
@@ -79,13 +67,53 @@ def test_gcs_estimator_checkpoint_save():
...
+    with patch(
+        "datasetinsights.storage.checkpoint.GCSEstimatorWriter",
+        MagicMock(return_value=mock_gcs_writer),
+    ):
+        writer = EstimatorCheckpoint._create_writer("gs://abucket/path", "def")
+
+        assert writer == mock_gcs_writer
+
+
+@patch("datasetinsights.storage.checkpoint.load_from_http")
```
adason: Why is mocking required as `_get_loader_from_path` only returns method? `load_local`, `load_from_gcs`, ` load_from_http` will not actually run. You should test directly if the desired method is returned by `_get_loader_from_path`

tests/simulation/test_download.py
```diff
@@ -54,13 +54,17 @@ def test_download_bad_request():
 
 def test_download_rows(downloader):
     n_rows = len(downloader.manifest)
-    with patch("datasetinsights.data.download.download_file") as mocked_dl:
+    with patch(
```
adason, line 57: Please merge from master. These linting are already fixed.

datasetinsights/storage/kfp_output.py
```diff
@@ -64,3 +64,9 @@ def write_metric(self):
             os.makedirs(self.filepath)
         with open(os.path.join(self.filepath, self.filename), "w") as f:
             json.dump(self.data, f)
+
+        logger.debug(
+            "Metrics file {0} saved at path: {1}".format(
```
adason: Use f-string in favor of `.format(...)`

datasetinsights/storage/checkpoint.py
```diff
@@ -162,32 +169,47 @@ def save(self, estimator, epoch=None):
...
+def load_from_http(estimator, url):
+    """ Load estimator from checkpoint files on GCS
+
+    Args:
+        estimator (datasetinsights.estimators.Estimator):
+            datasetinsights estimator object.
+        url: URL of the checkpoint file
+    """
+    temp_dir = tempfile.TemporaryDirectory().name
+    path = os.path.join(temp_dir, "filename")
```
adason: [nit] Might be better to user other names except "filename".

adason (changes requested):

datasetinsights/storage/checkpoint.py
Saurav-D: Since we are returning the path here, do we not want to keep the directory instead of deleting it?

datasetinsights/storage/checkpoint.py
Saurav-D: Fixed this.

adason (approved):

PR #9, opened 2020-07-27
ignore mock data and do isort for all files

BlairLee:
# Peer Review Information

do isort in the current master and ignore `mock_data` folder when doing `end-of-file-fixer` check.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,6 @@ repos:
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
-        exclude: ^mock_data/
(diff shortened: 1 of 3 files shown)


BlairLee:
Fixed a bug in `test_download.py`

.pre-commit-config.yaml
```diff
@@ -11,6 +11,7 @@ repos:
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
+        exclude: ^mock_data/
```
adason: We shouldn't ignore mock_data folder. There are files that should be checked (e.g. yaml file)

adason (approved):

PR #11, opened 2020-07-29
Readthedocs

AdamPalmarUnity:
The PR is for changes that goes into having readthedocs generate the documentation for the repo. 

The visualization module is currently not being generated in the documentation because it is causing the build agent to crash with a segmentation fault. 

The .rst files are generated from the Makefile.


datasetinsights/cli.py
```diff
@@ -179,8 +179,9 @@ def run(command, cfg):
     # todo this makes it so that we lose the tensorboard writer of non-master
     # processes which could make debugging harder
     writer = SummaryWriter(logdir, write_to_disk=is_master())
-    kfp_writer = KubeflowPipelineWriter(filename=cfg.system.metricsfilename,
-                                        filepath=cfg.system.metricsdir)
+    kfp_writer = KubeflowPipelineWriter(
```
adason: Pull from master as these are already fixed.

datasetinsights/data/datasets/__init__.py
```diff
@@ -6,11 +6,11 @@
...
 
 __all__ = [
-    Cityscapes,
-    CocoDetection,
-    Dataset,
-    GroceriesReal,
-    GoogleGroceriesReal,
-    NyuDepth,
-    SynDetection2D,
+    "Cityscapes",
```
adason, line 9: Why are these classes converted to strings?

docs/source/conf.py
```diff
@@ -3,24 +3,27 @@
...
-# sys.path.insert(0, os.path.abspath('../'))
 # -- Project information -----------------------------------------------------
 
 project = "datasetinsights"
 copyright = "2020, Unity Technologies"
 author = "Unity Technologies"
 
 # The full version, including alpha/beta/rc tags
-release = pkg_resources.get_distribution(project).version
+release = "0.1.0"
```
adason: [mustfix] I think it's not ideal to have to change this version for every tag release. Can you find out how to read package version automatically?

docs/source/conf.py
```diff
@@ -30,11 +33,16 @@
     "sphinx_rtd_theme",
     "sphinx.ext.napoleon",
 ]
+
+# autoapi_dirs = ['../../datasetinsights']
```
adason: Maybe we can remove this line as it was commented out?

pyproject.toml
```diff
@@ -85,6 +87,8 @@ exclude = '''
 )
 '''
 
+[tool.poetry.extras]
+docs = ["sphinx-rtd-theme", "recommonmark"]
```
adason, line 90: [nit] put a space between each block of pyproject.toml file.

pyproject.toml
```diff
@@ -44,6 +45,7 @@ tqdm = "^4.45.0"
 dask = {extras = ["complete"], version = "^2.14.0"}
 dash = "1.12.0"
 tensorflow = "^2.2.0"
+sphinx-autoapi = "^1.4.0"
```
adason: Is this package still required?

adason (changes requested):

datasetinsights/data/datasets/__init__.py
AdamPalmarUnity: https://docs.python.org/3/reference/simple_stmts.html#the-import-statement

docs/source/conf.py
AdamPalmarUnity: I will try adding the pkg_resource again as a module and see if it works now that imports are installed

docs/source/conf.py
AdamPalmarUnity: Wil do

pyproject.toml
AdamPalmarUnity: Ill double check it might not be needed anymore.

pyproject.toml
AdamPalmarUnity: done

datasetinsights/data/datasets/__init__.py
AdamPalmarUnity: The public names defined by a module are determined by checking the module’s namespace for a variable named __all__; if defined, it must be a sequence of strings which are names defined or imported by that module. The names given in __all__ are all considered public and are required to exist. If __all__ is not defined, the set of public names includes all names found in the module’s namespace which do not begin with an underscore character ('_'). __all__ should contain the entire public API. It is intended to avoid accidentally exporting items that are not part of the API (such as library modules which were imported and used within the module).

adason (dismissed):
You might want to run `poetry update` to make sure poetry.lock file is up to date with your changes in `pyproject.toml` during this review. Also, please update readthedocs to read from master branch once this review is merged 😸

datasetinsights/data/datasets/__init__.py
adason: I think it's fair to assume all of these module are public api. The only reason to hav __all__ is to prevent flake8 F401 [error](https://flake8.pycqa.org/en/2.6.0/warnings.html). We might as well remove __all__ statement and add #noqa string here.

AdamPalmarUnity: Do you want me to add the #noqa in this PR? 
Also not sure what #noqa is.

docs/source/conf.py
AdamPalmarUnity: It works now that the repo is installed properly

datasetinsights/data/datasets/__init__.py
adason: https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

adason: https://docs.python.org/3/tutorial/modules.html#importing-from-a-package

adason (approved):

PR #14, opened 2020-07-30
GitHub actions

AdamPalmarUnity:
# Peer Review Information

Added two workflows that will both trigger on release when published.
publish-docker-hub.yml builds and pushes the docker file to docker hub.
publish-pyp.yml installs poetry and builds the package and pushes it to pypi
publishing to the package was tested on testpypi.


.github/workflows/publish-docker-hub.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - name: Check out the repo
+        uses: actions/checkout@v2
+      - name: Push to Docker Hub
+        uses: docker/build-push-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
+          repository: unitytechnologies/datasetinsights
+          tag_with_ref: true
```
Saurav-D: Add new line at EOF to pass linting.

adason:
@AdamPalmarUnity A few minor things: 1. Since there are already a CI file called `main.yaml`, should we have a file called `release.yaml` for all tag release github actions? Maybe we should rename `main.yaml` to `linting-and-unittests.yaml` to make it consistent with `publish-docker-hub.yml` and `publish-pypi.yml`? 2. (this is really minor) Maybe we should use `.yaml` suffix to make it consistent with all yaml files in this repo? I know both `.yml` and `.yaml` are valid extensions.

Saurav-D (approved):

PR #16, opened 2020-07-30
Adding new Perception_Statistics.ipynb notebook 

JonathanHUnity:
# Peer Review Information
This PR adds a new public notebook called Perception_Statistics which contains stats and visualizations for the annotations and metrics supported by the Perception package out of the box. For the most part it follows the same style as the SynthDet notebook. One significant way it differs is that it supports each type of data being missing from the dataset, allowing it to be used in any configuration of Perception and not just when all labelers are present.

It also makes a few changes which fixes errors when loading a dataset that has no annotations.

Feel free to suggest where I should add tests or documentation

# Pull Request Check List


datasetinsights/data/simulation/captures.py
```diff
@@ -122,14 +123,20 @@ def _load_annotations(self, data_root, version):
...
+                    self.TABLE_NAME,
+                    version,
+                    record_path="annotations",
+                    meta="id",
+                    meta_prefix="capture.",
+                )
+            except KeyError:
+                annotation = pd.DataFrame(
+                    {"annotation_definition": [], "capture.id": []}
+                )
```
BlairLee: Can you add a unittest for this exception?

BlairLee:
With all those updates, are we going to rename the `SynthDet_Statistics.ipynb` to `Perception_Statistics.ipynb`?

JonathanHUnity:
> With all those updates, are we going to rename the `SynthDet_Statistics.ipynb` to `Perception_Statistics.ipynb`?

This is adding a new notebook that is more general to Perception. SynthDet_Statistics.ipynb has a bunch of statistics that are specific to SynthDet, so I think it's best to keep both. What do you think?

datasetinsights/data/simulation/captures.py
JonathanHUnity: Sure. do you want a test that `load_table` throws this exception in certain cases or a test that Captures(...) properly handles this case?

BlairLee: Would be better to test whether `Captures._load_annotations` can handle this case.

JonathanHUnity:
@BlairLee Where should we add documentation about the new notebook, if anywhere?

BlairLee:
> @BlairLee Where should we add documentation about the new notebook, if anywhere?

I believe it should be in the end-to-end documentation. (https://github.com/Unity-Technologies/SynthDet/pull/20/files)

BlairLee (approved):

PR #20, opened 2020-07-31
Report three metrics

BlairLee:
# Peer Review Information

Make faster-rcnn training report three evaluation metrics by default:
AP(IOU=50)
AP(IOU=0.5:0.95)
AR(IOU=0.5:0.95, max_detection=100)

Check the kubeflow runs:
[real2real model evaluation](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/pipeline/#/runs/details/84ccab51-7b9e-4f39-b264-e24c70a23545)
[finetuned model evaluation](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/pipeline/#/runs/details/2a922e17-5a24-4140-b101-45145f916eeb)
 

# Pull Request Check List


datasetinsights/estimators/faster_rcnn.py
```diff
@@ -377,7 +377,7 @@ def log_metric_val(self, label_mappings, epoch):
                 [result_per_label for result_per_label in result.values()]
             )
             logger.info(f"metric {metric_name} has mean result: {mean_result}")
-            self.writer.add_scalar(f"val/m{metric_name}", mean_result, epoch)
+            self.writer.add_scalar(f"val/{metric_name}", mean_result, epoch)
```
Saurav-D, line 380: Do we want to change the name for `kfp_metrics` as well? Can we instead change metric name instead of adding `m` prefix?

datasetinsights/evaluation_metrics/average_precision_2d_bbox.py
```diff
@@ -20,15 +20,21 @@ class AveragePrecisionBBox2D(EvaluationMetric):
...
 
     Args:
-        iou_threshold (float): iou threshold (default: 0.5)
+        iou_start (float): iou range starting point (default: 0.5)
+        iou_end (float): iou range ending point (default: 0.95)
+        iou_step (float): iou step size (default: 0.05)
         interpolation (string): AP interoperation method name for AP calculation
     """
 
     def __init__(
```
adason, line 28: This seems to be a design that needs to be improved. It's very confusing for user to "remember" using `iou_start=0.5, iou_end=0.95, iou_step=0.05` in order to compute "mAP@50IOU". Are there more intuitive interface? How can one provide an evaluation metric that support both "mAP@50IOU" and ""mAP[0.5:0.95]"? Do they have to share the same initialization method? Do they need to be the same class? Open-close principle?

datasetinsights/configs/faster_rcnn_groceries_real.yaml
```diff
@@ -27,15 +27,26 @@ test:
       version: v3
       split: test
 metrics:
-  AP:
+  mAP:
     name: AveragePrecisionBBox2D
     args:
-      iou_threshold: .50
+      iou_start: .50
```
adason: Is there a way that user does not have to remember these parameters? It's very common for object detection to support `mAP[.5:.95]`, `mAP@0.5` and `mAP[0.5:0.95]@100`. Maybe we can have these evaluation metric classes with default values?

tests/evaluation_metrics/test_average_precision_2d_bbox.py
```diff
@@ -17,7 +17,9 @@ def test_average_precision_2d_bbox(get_gt_pred_bbox):
     mini_batch3 = [[gt_bboxes[4], pred_bboxes[4]]]
 
     # test iou threshold = 0.5
-    ap_metrics = AveragePrecisionBBox2D(iou_threshold=0.5)
+    ap_metrics = AveragePrecisionBBox2D(
```
adason: Since the interface was changed, it forces you to modify unittests. Can changes to support mAP[.5:.95] be viewed as extension to your existing code, without changing the interface?

adason (changes requested):

datasetinsights/configs/faster_rcnn_groceries_real.yaml
BlairLee: Okay. Will make them default values.

datasetinsights/evaluation_metrics/average_precision_2d_bbox.py
BlairLee: I would create a new class for `mAP` and keep the original `AP` class.

tests/evaluation_metrics/test_average_precision_2d_bbox.py
BlairLee: Will do.

BlairLee:
Separate mAP and mAR with AP and AR. Add unittests for mAP and mAR. Cleanup the yaml file. Validate metrics on [kubeflow](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/pipeline/#/runs/details/58db44eb-2e87-4141-ab91-326173ba4e99).

datasetinsights/evaluation_metrics/average_precision_50iou.py
```diff
@@ -0,0 +1,18 @@
...
+
+https://github.com/rafaelpadilla/Object-Detection-Metrics#average-precision\
+Update algorithm from:
+https://github.com/rafaelpadilla/Object-Detection-Metrics/blob/master/lib/Evaluator.py
+"""
+
+from .average_precision_2d_bbox import AveragePrecisionBBox2D
+
+
+class AveragePrecision50IOU(AveragePrecisionBBox2D):
```
adason: Why use inheritance for `AveragePrecision50IOU` instead of composition?

datasetinsights/evaluation_metrics/mean_average_precision_50iou.py
```diff
@@ -0,0 +1,34 @@
+r"""Reference.
```
adason: I think it's overkill to have one file per evaluation metrics. Can you organize them in the same file so that it's easier to maintain these evaluation metrics? I think it make sense to have a file called `average_precision.py` which contains `mAP, `AP@50IOU`, ... and `average_recall.py` to include `mAR`, `AR[0.5:0.95]`, ...

To be more explicit and distinguish it from 3D, you can have  average_precision_2d.py and average_precision_3d.py, ... or anything that make sense to organize these evaluation metrics.

datasetinsights/evaluation_metrics/mean_average_precision_50iou.py
```diff
@@ -0,0 +1,34 @@
...
+Update algorithm from:
+https://github.com/rafaelpadilla/Object-Detection-Metrics/blob/master/lib/Evaluator.py
+"""
+
+import numpy as np
+
+from .average_precision_2d_bbox import AveragePrecisionBBox2D
+
+
+class MeanAveragePrecision50IOU(AveragePrecisionBBox2D):
```
adason: Why use inheritance over composition?

datasetinsights/evaluation_metrics/base.py
```diff
@@ -31,14 +31,26 @@ def find(name):
...
+        metric_classes = EvaluationMetric.get_all_subclasses(EvaluationMetric)
         metric_names = [d.__name__ for d in metric_classes]
         if name in metric_names:
             metric_cls = metric_classes[metric_names.index(name)]
             return metric_cls
         else:
             raise NotImplementedError(f"Unknown Metric class {name}!")
 
+    @staticmethod
+    def get_all_subclasses(cls):
```
adason: This can be prevented if we favor composition over inheritance.

datasetinsights/evaluation_metrics/mean_average_recall.py
```diff
@@ -0,0 +1,81 @@
...
+        iou_start (float): iou range starting point (default: 0.5)
+        iou_end (float): iou range ending point (default: 0.95)
+        iou_step (float): iou step size (default: 0.05)
+        max_detections (int): max detections per image (default: 100)
+    """
+
+    def __init__(
+        self, iou_start=0.5, iou_end=0.95, iou_step=0.05, max_detections=100
+    ):
+        self.iou_thresholds = np.linspace(
```
adason: [nit] why is this instance variable required. why should this variable be public? Can it be private instance variable?

datasetinsights/evaluation_metrics/mean_average_recall.py
```diff
@@ -0,0 +1,81 @@
...
+            iou_start,
+            iou_end,
+            np.round((iou_end - iou_start) / iou_step) + 1,
+            endpoint=True,
+        )
+        self.max_detections = max_detections
+        self.mar_records = [
```
adason: [nit] mar_records are a little too generic. Maybe `mar_per_iou`?

datasetinsights/evaluation_metrics/mean_average_recall.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class MeanAverageRecall(EvaluationMetric):
+    """2D Bounding Box Mean Average Recall metrics.
+
+    Implementation of classic mAR metrics. We use 10 IoU thresholds
+    of .50:.05:.95. This is the same metrics in cocoEval.summarize():
+    Average Recall (AR) @[IoU=0.50:0.95 | area = all | maxDets=100]
+
+    Attributes:
```
adason: Are these attributes required to be public? Do you need users to use these instance variables?

datasetinsights/evaluation_metrics/mean_average_recall.py
```diff
@@ -0,0 +1,81 @@
...
+        )
+        self.max_detections = max_detections
+        self.mar_records = [
+            AverageRecallBBox2D(iou, max_detections)
+            for iou in self.iou_thresholds
+        ]
+
+    def reset(self):
+        """Reset metrics."""
+        self.mar_records = [
```
adason: [nit] might be simpler call `[r.reset() for r in self.mar_records]`

datasetinsights/evaluation_metrics/mean_average_recall.py
```diff
@@ -0,0 +1,81 @@
...
+            where gt_bboxes1, pred_bboxes1 contain gt bboxes and pred bboxes
+            in one image
+        """
+        for mean_ar in self.mar_records:
+            mean_ar.update(mini_batch)
+
+    def compute(self):
+        """Compute AR for each label.
+
+        Return:
```
adason: Have you build the documentations to see what the docstring looks like? Is this the proper return format?

tests/evaluation_metrics/test_average_precision_2d_bbox.py
```diff
@@ -1,9 +1,16 @@
...
+from datasetinsights.evaluation_metrics import (
+    AveragePrecision50IOU,
+    AveragePrecisionBBox2D,
+    MeanAveragePrecision,
+    MeanAveragePrecision50IOU,
+)
 
 
-def test_average_precision_2d_bbox(get_gt_pred_bbox):
+def get_mini_batches(get_gt_pred_bbox):
```
adason: Maybe this can be a pytest fixture?

tests/evaluation_metrics/test_average_precision_2d_bbox.py
```diff
@@ -32,3 +43,94 @@ def test_average_precision_2d_bbox(get_gt_pred_bbox):
...
+):
+    mini_batches = get_mini_batches(get_gt_pred_bbox)
+
+    map_metrics = MeanAveragePrecision50IOU()
+    for mini_batch in mini_batches:
+        map_metrics.update(mini_batch)
+
+    map_metrics.compute()
+    assert mock_update.call_count == len(mini_batches)
+    mock_compute.assert_called()
```
adason: Why not make `mock_compute` to return some data so that you can compute the result against `map_metrics.compute()`? Otherwise, this test is not very strong test as there are many way to pass this test and produce incorrect result.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -373,27 +371,25 @@ def log_metric_val(self, label_mappings, epoch):
...
             logger.debug(result)
-            mean_result = np.mean(
-                [result_per_label for result_per_label in result.values()]
-            )
-            logger.info(f"metric {metric_name} has mean result: {mean_result}")
-            self.writer.add_scalar(f"val/m{metric_name}", mean_result, epoch)
-
-            self.kfp_writer.add_metric(name=metric_name, val=mean_result)
+            logger.info(f"metric {metric_name} has result: {result}")
+            if not isinstance(result, dict):
```
adason: This has very implicit assumption about types of the metrics that was computed. What if the metrics now return a list instead of dict? Can this be improved? 

Since you get full control of the evaluation metrics (through the yaml file), can you know if each metric produces a single number or a dict of {string: float} mappings?

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+        interpolation="EveryPointInterpolation",
+    ):
+        self.iou_thresholds = np.linspace(
+            iou_start,
+            iou_end,
+            np.round((iou_end - iou_start) / iou_step) + 1,
+            endpoint=True,
+        )
+        self.interpolation = interpolation
+        self.map_records = [
```
adason: Same as comments in `MeanAverageRecall`.

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+            AveragePrecisionBBox2D(iou, interpolation)
+            for iou in self.iou_thresholds
+        ]
+
+    def reset(self):
+        """Reset metrics."""
+        self.map_records = [
```
adason: Same as comments in `MeanAverageRecall`.

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+            mean_ap.update(mini_batch)
+
+    def compute(self):
+        """Compute AP for each label.
+
+        Return:
+            mAP (float): mean average precision across all ious
+        """
+        mean_sum = 0
+        for mean_ap in self.map_records:
```
adason: This violates DRY. The logic of average over a range of IOUs are repeated in multiple places. Can you improve this and create shared modules?

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+from .average_precision_2d_bbox import AveragePrecisionBBox2D
+from .base import EvaluationMetric
+
+
+class MeanAveragePrecision(EvaluationMetric):
+    """2D Bounding Box Mean Average Precision metrics.
+
+    Implementation of classic mAP metrics. We use 10 IoU thresholds
+    of .50:.05:.95. This is the same metrics in cocoEval.summarize():
+    Average Precision (AP) @[IoU=0.50:0.95 | area = all | maxDets=100]
```
adason: Do we limit the maxDets to 100 in this evaluation metrics? It's not clear to me (and users) if this was implemented.

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+
+from .average_precision_2d_bbox import AveragePrecisionBBox2D
+from .base import EvaluationMetric
+
+
+class MeanAveragePrecision(EvaluationMetric):
+    """2D Bounding Box Mean Average Precision metrics.
+
+    Implementation of classic mAP metrics. We use 10 IoU thresholds
+    of .50:.05:.95. This is the same metrics in cocoEval.summarize():
```
adason: [nit] Should probable change this to 0.5:0.95:0.0.5. It's common to see this range in `start:stop:step` format.

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+    Average Precision (AP) @[IoU=0.50:0.95 | area = all | maxDets=100]
+
+    Attributes:
+        map_records (dict): save prediction records for each ious
+        iou_thresholds (numpy.array): iou thresholds
+
+    Args:
```
adason: Args is not consistent with the actual args here.

datasetinsights/evaluation_metrics/mean_average_precision.py
```diff
@@ -0,0 +1,83 @@
...
+import numpy as np
+
+from .average_precision_2d_bbox import AveragePrecisionBBox2D
+from .base import EvaluationMetric
+
+
+class MeanAveragePrecision(EvaluationMetric):
+    """2D Bounding Box Mean Average Precision metrics.
+
+    Implementation of classic mAP metrics. We use 10 IoU thresholds
```
adason: I would rephrase this to be more explicit. Suggestion:
This implementation computes Mean Average Precision (mAP) metric, which is implemented as the Average Precision average over all labels and IOU thresholds [0.5:0.95:0.05]. 

mAP = \mean_{label, IOU} AP(label, IOU)

adason (changes requested):
The documentation of related evaluation metrics `AveragePrecision`, `MeanAveragePrecisionAverageOverIOU`, `MeanAveragePrecisionIOU50`, `AverageRecall`, `MeanAverageRecallAverageOverIOU`, `MeanAverageRecall` should be greatly improved to indicate what each metrics tries to compute. There should be **absolutely** no ambiguity here. Use math formula might help if this is harder to explain in words.

datasetinsights/evaluation_metrics/base.py
BlairLee: Agree. Now I think the composition is better.

datasetinsights/evaluation_metrics/mean_average_recall.py
BlairLee: fixed.

datasetinsights/evaluation_metrics/mean_average_recall.py
BlairLee: Set it as private instance variable.

datasetinsights/estimators/faster_rcnn.py
BlairLee: How about we use `metric_name.startswith("m")` to check whether is a single number or dict? We know in 2d object detection, metric name starting with "m", means the mean value, like mAP, mAR, etc.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -373,27 +371,25 @@ def log_metric_val(self, label_mappings, epoch):
...
             logger.debug(result)
-            mean_result = np.mean(
-                [result_per_label for result_per_label in result.values()]
-            )
-            logger.info(f"metric {metric_name} has mean result: {mean_result}")
-            self.writer.add_scalar(f"val/m{metric_name}", mean_result, epoch)
-
-            self.kfp_writer.add_metric(name=metric_name, val=mean_result)
+            logger.info(f"metric {metric_name} has result: {result}")
+            if metric_name.startswith("m"):
```
adason: [mustfix] This assumes knowledge "mXX" are single value metric. Can you directly add this attribute in `EvaluationMetric` class? In this way it's very explicit and does not break the pipeline if users decide to change their metric name alias from `mAP` to `MAP` or something else.

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -196,3 +185,105 @@ def n_point_interpolated_ap(recall, precision, point=11):
...
 
         return ap
+
+
+class AveragePrecisionIOU50(EvaluationMetric):
+    """2D Bounding Box Average Precision at IOU = 50%.
+
+    This implementation would calculate AP@50IOU for each label.
+
+    Attributes:
```
adason: [nit] The attributes ap here seems to suggests `AveragePrecisionIOU50.ap` and `MeanAveragePrecisionIOU50.ap` store the same info. They are not (I think). Maybe use slightly different name might be better. 

Also, should this a public instance variable?

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -196,3 +185,105 @@ def n_point_interpolated_ap(recall, precision, point=11):
...
+        self.ap.update(mini_batch)
+
+    def compute(self):
+        return self.ap.compute()
+
+
+class MeanAveragePrecisionIOU50(EvaluationMetric):
+    """2D Bounding Box Mean Average Precision metrics at IOU=50%.
+
+    This implementation would calculate mAP@50IOU. The metric can be
```
adason: Remove `The metric can be describe...`. Simply have the formula is enough.

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -196,3 +185,105 @@ def n_point_interpolated_ap(recall, precision, point=11):
...
+    described as:
+    mAP = mean_{label}AP(label)@IOU50
```
adason: Have you checked documentation to see if this formula is rendered properly?

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -196,3 +185,105 @@ def n_point_interpolated_ap(recall, precision, point=11):
...
+        return mean_ap
+
+
+class MeanAveragePrecisionAverageOverIOU(EvaluationMetric):
+    """2D Bounding Box Mean Average Precision metrics.
+
+    This implementation computes Mean Average Precision (mAP) metric,
+    which is implemented as the Average Precision average over all
+    labels and IOU thresholds [0.5:0.95:0.05]. The max detections
+    per image is limited to 100. The metric can be described as:
```
adason: Remove `The metric can be described as:`

datasetinsights/evaluation_metrics/average_recall_2d.py
```diff
@@ -89,26 +92,58 @@ def compute(self):
...
+        """
+        for mean_ar in self.mar_per_iou:
+            mean_ar.update(mini_batch)
+
+    def compute(self):
+        """Compute AR for each label.
 
-        Returns:
-            labels (dict): a dictionary of prediction boundign boxes
+        Returns (float):
```
adason: Have you checked if the document is rendered correctly?

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
+import collections
+
+import numpy as np
+
+
+def mean_metrics_over_iou(metric_per_iou):
```
adason: Did you add unittests for this method?

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    mean_sum = 0
+    for mean_ap in metric_per_iou:
+        result = mean_ap.compute()
+        mean_sum += np.mean(
+            [result_per_label for result_per_label in result.values()]
+        )
+    return mean_sum / len(metric_per_iou)
+
+
+def filter_pred_bboxes(pred_bboxes, max_detections):
```
adason: Dod you add unittests for this method?

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
+import collections
+
+import numpy as np
+
+
+def mean_metrics_over_iou(metric_per_iou):
+    """Calculate mean value over IOUs.
+
+    Args:
+        metric_per_iou (dict): metric records for each iou
```
adason: This method is more general than `average_over_iou`. It can apply to any metrics that return a dict of values such as `{label: float}`. Maybe we should have a general method that average over dict of values?

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
...
+
+    Returns (float):
+        mean metric values over IOUs
+    """
+    mean_sum = 0
+    for mean_ap in metric_per_iou:
+        result = mean_ap.compute()
```
adason: Why is `metric.compute` called here? This is not the responsibility of the `mean` operation, is it?

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
...
+        mean_sum += np.mean(
+            [result_per_label for result_per_label in result.values()]
+        )
+    return mean_sum / len(metric_per_iou)
+
+
+def filter_pred_bboxes(pred_bboxes, max_detections):
+    """Save bboxes with same label in to a dictionary.
+
+    This operation only apply to predictions for a single image.
```
adason: I am confused by what this method tries to achieve. The input is a list of bounding box object, and the output is also a list of bounding boxes, right? So why is the concept `labels` applies to this method?

tests/evaluation_metrics/conftest.py
```diff
@@ -4,7 +4,7 @@
 
 
 @pytest.fixture
-def get_gt_pred_bbox():
+def get_mini_batches():
```
adason: Why not create another fixture so that you don't have to change `test_confusion_metrix.py`?

datasetinsights/evaluation_metrics/average_recall_2d.py
```diff
@@ -89,26 +92,58 @@ def compute(self):
...
-        """Save bboxes with same label in to a dictionary.
 
-        This operation only apply to predictions for a single image.
+class MeanAverageRecallAverageOverIOU(EvaluationMetric):
+    """2D Bounding Box Mean Average Recall metrics.
+
+    This implementation computes Mean Average Recall (mAR) metric,
+    which is implemented as the Average Recall average over all
+    labels and IOU thresholds [0.5:0.95:0.05]. The max detections
+    per image is limited to 100. The metric can be described as:
```
adason: Remove `The metric can be described as`

datasetinsights/evaluation_metrics/average_recall_2d.py
```diff
@@ -89,26 +92,58 @@ def compute(self):
...
+    """
+
+    IOU_THRESHOULDS = np.linspace(
+        0.5, 0.95, np.round((0.95 - 0.5) / 0.05) + 1, endpoint=True
+    )
+
+    def __init__(self):
+        self.mar_per_iou = [
+            AverageRecall(iou)
+            for iou in MeanAverageRecallAverageOverIOU.IOU_THRESHOULDS
```
adason, line 114: [nit] Can use `self.IOU_THRESHOULDS` to save some typing

datasetinsights/evaluation_metrics/metrics_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    for mean_ap in metric_per_iou:
+        result = mean_ap.compute()
+        mean_sum += np.mean(
+            [result_per_label for result_per_label in result.values()]
+        )
+    return mean_sum / len(metric_per_iou)
+
+
+def filter_pred_bboxes(pred_bboxes, max_detections):
+    """Save bboxes with same label in to a dictionary.
```
adason: The title string is not clear what this method is trying to achieve. Also, it should be super for users that `max_detection` applies to capped at the "top 100 detections ordered by decreasing detection score".

adason (changes requested):

datasetinsights/estimators/faster_rcnn.py
BlairLee: Okay. Will add it as an attribute in `EvaluationMetric` class.

datasetinsights/evaluation_metrics/average_precision_2d.py
BlairLee: They both use `AveragePrecision` to calculate Average Precision at IOU = 50%. The only difference is in the `compute()` method. 
If they are not public instance variables, do you want to make them as class attributes?

adason: I see the differences between both classes. It seems `AveragePrecisionIOU50` is simple `AveragePrecision` with some default parameter. Why not use `__getattr__` to prevent rewriting `compute`, `reset`, `update` method?

datasetinsights/evaluation_metrics/metrics_utils.py
BlairLee: Will move `compute()` out of this method.

datasetinsights/evaluation_metrics/average_precision_2d.py
BlairLee: I don't think `__getattr__` would work. Going to stick with the previous version.

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -8,47 +8,44 @@
...
     """2D Bounding Box Average Precision metrics.
-
-    Attributes:
-        label_records (dict): save prediction records for each label
-        ap_method (string): AP interoperation method name for AP calculation
-        {"EveryPointInterpolation"| "NPointInterpolatedAP"}
-        gt_bboxes_count (dict): ground truth box count for each label
-        iou_threshold (float): iou threshold
-
-    Args:
```
adason, line 25: Fix Args docstring here.

datasetinsights/io/bbox.py
```diff
@@ -4,6 +4,25 @@
 from pyquaternion import Quaternion
 
 
+def group_bbox2d_per_label(bboxes):
```
adason, line 7: Excellent!!!

datasetinsights/evaluation_metrics/__init__.py
```diff
@@ -12,8 +17,12 @@
     IoU,
```
adason: __all__ should be a list of strings instead of list of class names. 

datasetinsights/evaluation_metrics/average_precision_2d.py
adason: Should add a space an `.. math::` flag for math formula to be parsed by latex properly.

adason (dismissed):

sanjayuconn (approved):

datasetinsights/evaluation_metrics/average_recall_2d.py
```diff
@@ -0,0 +1,133 @@
...
+            in one image
+        """
+        for bboxes in mini_batch:
+            gt_bboxes, pred_bboxes = bboxes
+            for gt_bbox in gt_bboxes:
+                self._gt_bboxes_count[gt_bbox.label] += 1
+
+            pred_bboxes = sorted(
+                pred_bboxes, key=lambda bbox: bbox.score, reverse=True
+            )[: self._max_detections]
```
adason, line 60: @BlairLee This will fail if `self._max_detections < len(pred_bboxes)`. Also, this behavior should be well documented in `AverageRecall` class docstring.

BlairLee: Actually, if `self._max_detections < len(pred_bboxes)`, it would keep all the boxes and no error would be raised. Will add docstring.

PR #22, opened 2020-08-04
Refactor datasetinsights CLI

adason:
# Peer Review Information

This change introduces new CLI for the datasetinsights package using [click](https://click.palletsprojects.com/en/7.x/). This change only includes the interface and the actual implementation for `download`, `train` and `evaluate` are left for subsequence reviews. 

# Pull Request Check List


datasetinsights/commands/download.py
```diff
@@ -0,0 +1,106 @@
...
+    "-n",
+    "--name",
+    type=click.STRING,
+    required=True,
+    help="The dataset registry name.",
+)
+@click.option(
+    "-d",
+    "--data-root",
+    type=click.STRING,
```
AdamPalmarUnity: Should be changed to `click.Path(exists=True)` instead of click.STRING this way we can avoid doing all the validation work

datasetinsights/commands/evaluate.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@click.command(
+    help="Start model evaluation tasks for a pre-trained model.",
+    context_settings=const.CONTEXT_SETTINGS,
+)
+@click.option(
+    "-c",
+    "--config",
+    type=click.STRING,
```
AdamPalmarUnity, line 17: Can be changed to `click.File()`

datasetinsights/commands/evaluate.py
```diff
@@ -0,0 +1,91 @@
...
+    "-p",
+    "--checkpoint-file",
+    type=click.STRING,
+    required=True,
+    help="URI to a checkpoint file.",
+)
+@click.option(
+    "-d",
+    "--data-root",
+    type=click.STRING,
```
AdamPalmarUnity: Can be changed to `click.Path(exists=True)`

datasetinsights/commands/evaluate.py
```diff
@@ -0,0 +1,91 @@
...
+        "Number of multiprocessing workers for loading datasets. "
+        "Set this argument to 0 will disable multiprocessing which is "
+        "recommended when running inside a docker container."
+    ),
+)
+@click.option(
+    "--kfp-metrics-dir",
+    type=click.STRING,
+    default=const.DEFAULT_KFP_METRICS_DIR,
+    help="Path to the directory where Kubeflow Metrics files are stored.",
```
AdamPalmarUnity, line 61: Is this local only or will a GCS URI also work? If its a local file only it makes sense to make the type `click.Path(exists=True)`

datasetinsights/commands/train.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@click.command(
+    help="Start model training (and optionally validation) tasks.",
+    context_settings=const.CONTEXT_SETTINGS,
+)
+@click.option(
+    "-c",
+    "--config",
+    type=click.STRING,
```
AdamPalmarUnity, line 17: Can be changed to `click.File()`

datasetinsights/commands/train.py
```diff
@@ -0,0 +1,99 @@
...
+    default=None,
+    help=(
+        "URI to a checkpoint file. If specified, model will load from "
+        "this checkpoint and resume training."
+    ),
+)
+@click.option(
+    "-d",
+    "--data-root",
+    type=click.STRING,
```
AdamPalmarUnity: Can be changed to `click.Path(exists=True)`

AdamPalmarUnity (changes requested):

datasetinsights/commands/download.py
```diff
@@ -0,0 +1,106 @@
+import logging
+
+import click
+
+import datasetinsights.constants as const
+
+logger = logging.getLogger(__name__)
+
+
+def validate_source_uri(ctx, param, value):
```
AdamPalmarUnity: This could be replaced with the extension of a click.ParamType class.
example

```
class SourceURI(click.ParamType):

    name = "uri"

    def convert(self, value, param, ctx):
        if os.path.exists(value) or value.startswith("gs://"):
            return value
        else:
            self.fail(f"The uri {value} has to exist on the local filesystem or be a uri to gcs ",
                      param,
                      ctx)
```

datasetinsights/commands/download.py
adason: Good suggestion. I add `click.Path(exists=True)` for all `data-root` CLI options.

datasetinsights/commands/evaluate.py
adason: `click.File()` will open the file by default which is usually less desirable. Also, we might want to support this config file from public http url (such as github repo) to allow users passing config files to docker container (instead of having these files that was included during the build process. 

I think it's better to keep it as is for now.

datasetinsights/commands/evaluate.py
adason: This will more likely only support local directory. I will make this `click.Path`.

datasetinsights/commands/evaluate.py
adason: fixed.

datasetinsights/commands/train.py
adason: sams as in evaluate.py. Will not change this to click.File()

datasetinsights/commands/train.py
adason: fixed.

datasetinsights/commands/download.py
adason: Good call. I will incorporate this suggestion.

datasetinsights/commands/evaluate.py
AdamPalmarUnity: I see that it might make more sense to make it a custom ParamType for the future

AdamPalmarUnity (approved):

PR #23, opened 2020-08-04
SynthDet Public URL Download Refactor

Saurav-D:
# Peer Review Information

- Refactored for download subset dataset from public url.
- The training task will depend on this step to make sure SynthDet data is available in the local file system.


datasetinsights/scripts/synthdet_download.py
```diff
@@ -0,0 +1,97 @@
...
+    else:
+        logger.info(
+            f"No dataset_uri is provided. Assuming the data root directory"
+            f" {data_root} already contains synthetic dataset."
+        )
+
+
+if __name__ == "__main__":
+    args = parse_args()
+    run(args)
```
BlairLee: [mustfix] It's better to refactor the `run` method, followed by this [implementation](https://github.com/Unity-Technologies/datasetinsights/pull/5/files#diff-8f2d8e40354b54404d275b10fbee6864R42), line 42-46 in `public_download.py`? Basically, it would be better to move download and unzip function, along with logging, in `synthetic.py` and make them static. In here, you can simply call `SynDetection2D.download()`

datasetinsights/configs/faster_rcnn_synthetic.yaml
```diff
@@ -7,6 +7,7 @@ train:
     name: SynDetection2D
     args:
       run_execution_id: zMvGEV0
+      load_local: False
```
BlairLee: Why do we need this option for users? Can we simply check if the dataset downloaded first? If not, we can download the synthetic data.

kubeflow/synthdet_train.py
```diff
@@ -0,0 +1,88 @@
...
+            f"--logdir={logdir}",
+            "train.epochs",
+            epochs,
+            "train.dataset.args.load_local",
+            True,
+            "val.dataset.args.load_local",
+            True,
+        ],
+        # Refer to pvloume in previous step to explicitly call out dependency
+        pvolumes={"/data": download.pvolumes["/data"]},
```
BlairLee: Can we record `mlpipeline-metrics` at the end of the training process?

datasetinsights/scripts/public_download.py
```diff
@@ -34,20 +41,37 @@ def parse_args():
...
 
 
-def download(name, data_root, version):
+def download(name, source_uri, data_root, version):
     # TODO this method should be refactored once we have clean download
     # CLI interface.
-    dataset = Dataset.find(name)
-    dataset.download(data_root, version)
+    if name.lower() == "Synthetic".lower():
+        SynDetection2D.download(source_uri, data_root)
```
BlairLee: Do we need to add source_uri in cli for Synthetic data? In GroceriesReal, we just hard coded the source_uri. It would be better to keep it consistent for both synthetic and real datasets. In that case, we can set `--name SynDetection2D`, and use `dataset = Dataset.find(name)` `dataset.download(source_data_root)`

datasetinsights/data/datasets/synthetic.py
```diff
@@ -297,10 +307,57 @@ def exists(capture_file):
...
+    def download_captures_from_manifest(self, manifest_file):
         """ Download captures of a given manifest file.
 
         Args:
             manifest_file (str): path to a manifest file
         """
         _download_captures(self.root, manifest_file)
+
+    @staticmethod
+    def download(source_uri, destination):
```
BlairLee: Can you add a checksum for the public synthetic subset? You can use `from datasetinsights.data.download import validate_checksum`, which is well tested. Also, take a look at `_download_http` method in `groceries_real.py` as an example.

datasetinsights/scripts/public_download.py
Saurav-D: The main reason for adding `source-uri` is that if user wants to use own dataset.

Saurav-D: But still we would need this because we don't pass version to Synthetic dataset as of now. Can we make synthetic version as `5k`, `10k` etc which are essentially according  to the size of the dataset?

BlairLee: Maybe we can name the synthetic dataset version as v1, v2. And write a doc talking about what those versions mean, such as "v1: 5k; v2: 10k". And use a name table to save public uri as well as checksum. Something like this in groceries_real.py:
```
GROCERIES_REAL_DATASET_TABLES = {
        "v3": GroceriesRealTable(
            "v3", "v3.zip", f"{PUBLIC_GROCERIES_REAL_PATH}/v3.zip", 2322380006,
        ),
    }
```

datasetinsights/data/datasets/synthetic.py
adason: @Saurav-D  Agree with @BlairLee. It's important (and typical) to provide checksums for large files. We should have simple checksum mechanism to make sure these files are not corrupted during the download process.

Saurav-D: @adason I've implemented the checksum, I will push the changes along with others. But I just wanted to know if we should make version of Synthdet dataset as describe in above comments.

datasetinsights/data/datasets/synthetic.py
```diff
@@ -297,10 +340,74 @@ def exists(capture_file):
...
+    def download_captures_from_manifest(self, manifest_file):
         """ Download captures of a given manifest file.
 
         Args:
             manifest_file (str): path to a manifest file
         """
         _download_captures(self.root, manifest_file)
+
+    @staticmethod
+    def download(data_root, version):
```
BlairLee: Can you add unittest for this download method? Make sure it raises exceptions correctly. You can take `test_groceriesreal.py` as an example.

datasetinsights/data/datasets/synthetic.py
```diff
@@ -297,10 +340,74 @@ def exists(capture_file):
...
+        """Downloads dataset zip file and unzips it.
+
+        Args:
+            data_root (str): Path where to download the dataset.
+            version (str): version of GroceriesReal dataset, e.g. "v1"
+
+        Note: Synthetic dataset is downloaded and unzipped to
+        data_root/synthetic.
+        """
```
BlairLee: Please specify what exceptions would raise in the docstrings? Like:
```
Raises:
     ValueError if the dataset version is not supported
     ChecksumError if the download file checksum does not match
     DownloadError if the download file failed
```

BlairLee (approved):

PR #25, opened 2020-08-06
Refactor CLI for train and evaluate

AdamPalmarUnity:
# Peer Review Information

Implementation of the cli commands datasetinsights train/evaluate

They are called the following way

Training
datasetinsights train --config config.yaml --train-data path-to-directory

Evaluate
datasetinsights evaluate --config config.yaml --test-data path-to-directory

# Pull Request Check List


datasetinsights/estimators/builder.py
```diff
@@ -0,0 +1,101 @@
...
+
+import datasetinsights.constants as const
+from datasetinsights.estimators import Estimator
+from datasetinsights.estimators.faster_rcnn import FasterRCNNDepedencies
+from datasetinsights.storage.checkpoint import EstimatorCheckpoint
+from datasetinsights.storage.kfp_output import KubeflowPipelineWriter
+from datasetinsights.torch_distributed import init_distributed_mode, is_master
+
+
+class EstimatorBuilder:
```
adason: What's the main reason to have an` EstimatorBuilder`. What problem is this trying to solve?

datasetinsights/estimators/builder.py
```diff
@@ -0,0 +1,101 @@
...
+    kfp_writer = KubeflowPipelineWriter(
+        filename=const.DEFAULT_KFP_METRICS_FILENAME,
+        filepath=const.DEFAULT_KFP_METRICS_DIR,
+    )
+    checkpointer = EstimatorCheckpoint(
+        estimator_name=model_config.estimator,
+        log_dir=writer.logdir,
+        distributed=distributed,
+    )
+    estimator_depedencies = FasterRCNNDepedencies(
```
adason: Why is this specific to `FasterRCNNDepedencies` here? Is `create_estomator_depedencies` expected to work for all estimators?

AdamPalmarUnity: This is my bad it should not be specific to FasterRCNNDepdencies. Thanks for catching that.

datasetinsights/estimators/builder.py
AdamPalmarUnity: The purpose is to reuse this code used to configure the evaluate and train estimator

```
    if len(ctx.args) == 1:
        model_config.merge_from_list(ctx.args[0].split(" "))
    else:
        model_config.merge_from_list(ctx.args)

    gpu, rank, distributed = init_distributed_mode()
    if torch.cuda.is_available() and not ctx.params["no_cuda"]:
        device = torch.device("cuda")
    else:
        device = torch.device("cpu")
    logdir = ctx.params["tb_log_dir"]
    if logdir == const.NULL_STRING:
        # Use logdir=None to force using SummaryWriter default logdir,
        # which points to ./runs/<model>_<timestamp>
        logdir = None

    # todo this makes it so that we lose the tensorboard
    #  writer of non-master processes which could make debugging harder
    writer = SummaryWriter(
        logdir,
        write_to_disk=is_master(),
        max_queue=const.SUMMARY_WRITER_MAX_QUEUE,
        flush_secs=const.SUMMARY_WRITER_FLUSH_SECS,
    )
    kfp_writer = KubeflowPipelineWriter(
        filename=const.DEFAULT_KFP_METRICS_FILENAME,
        filepath=const.DEFAULT_KFP_METRICS_DIR,
    )
    checkpointer = EstimatorCheckpoint(
        estimator_name=model_config.estimator,
        log_dir=writer.logdir,
        distributed=distributed,
    )
```

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -80,59 +141,60 @@ def __init__(
...
     def train(self, **kwargs):
         """start training, save trained model per epoch."""
         config = self.config
-        train_dataset = create_dataset(config, TRAIN)
-        val_dataset = create_dataset(config, VAL)
+        train_dataset = create_dataset(config, self.data_root, TRAIN)
+        val_dataset = create_dataset(config, self.data_root, VAL)
         label_mappings = train_dataset.label_mappings
-        if self.config.system.dryrun:
-            train_dataset = create_dryrun_dataset(config, train_dataset, TRAIN)
```
sanjayuconn, line 111: Why are you removing dry run? if we are not supporting dry run why we have create_dryrun_dataset method in this class?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -263,30 +325,29 @@ def train_one_epoch(
     def evaluate(self, **kwargs):
         """evaluate given dataset."""
         config = self.config
-        test_dataset = create_dataset(config, TEST)
+        test_dataset = create_dataset(config, self.data_root, TEST)
         label_mappings = test_dataset.label_mappings
-        if self.config.system.dryrun:
-            test_dataset = create_dryrun_dataset(config, test_dataset, TEST)
-
```
sanjayuconn, line 270: Why are you removing dry run? if we are not supporting dry run why we have create_dryrun_dataset method in this class?

tests/test_faster_rcnn.py
```diff
@@ -99,15 +104,20 @@ def test_faster_rcnn_train_all(mock_create, mock_loss, config, dataset):
     estimator = FasterRCNN(
         config=config,
         writer=writer,
-        device=torch.device("cpu"),
         checkpointer=checkpointer,
         kfp_writer=kfp_writer,
+        logdir="/tmp",
```
sanjayuconn, line 109: 1. Can we have one string dir="/tmp" instead of creating multiple "/tmp"?
2. After test are you deleting this dir?

datasetinsights/commands/train.py
```diff
@@ -96,4 +97,14 @@ def cli(
     ctx = click.get_current_context()
     logger.debug(f"Called train command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    # TODO: Call train command here.
+
+    from datasetinsights.estimators.factory import EstimatorFactory
```
adason: Please move this import to top of the file. We don't have to optimize for script launch speed as model training/evaluation usually takes a lot longer.

datasetinsights/estimators/deeplab.py
```diff
@@ -94,25 +94,24 @@ class DeeplabV3(Estimator):
...
-    def __init__(self, *, config, writer, checkpointer, device, **kwargs):
+    def __init__(self, config, writer, checkpointer, device, **kwargs):
         self.config = config
+        self.writer = writer
+        self.checkpointer = checkpointer
+        self.device = device
 
-        self.backbone = config.backbone
-        self.num_classes = config.num_classes
+        self.backbone = self.config.backbone
```
adason: [nit] why not just read from `config` variable directly which is in scope of `__init__` method?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
+from tensorboardX import SummaryWriter
```
adason: [nit] Since `EstimatorFactory` and `Estimator` are two of the classes that was coupled, it might be better to have them in the same file so that developers does not have to jump between two files?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
         config,
-        writer,
-        kfp_writer,
-        checkpointer,
-        device,
+        logdir,
         box_score_thresh=0.05,
-        gpu=0,
-        rank=0,
+        data_root=None,
```
sanjayuconn: Why logdir and data_root is not part of config?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+
+import datasetinsights.constants as const
```
sanjayuconn: Can we have class level docstring?

datasetinsights/estimators/deeplab.py
```diff
@@ -94,25 +94,24 @@ class DeeplabV3(Estimator):
         lr_scheduler: pytorch learning rate scheduler
     """
 
-    def __init__(self, *, config, writer, checkpointer, device, **kwargs):
```
sanjayuconn, line 97: Why are we removing *?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+from datasetinsights.estimators import Estimator
+from datasetinsights.storage.checkpoint import EstimatorCheckpoint
+from datasetinsights.storage.kfp_output import KubeflowPipelineWriter
+
+
+class EstimatorFactory:
+    @staticmethod
+    def create(name, **kwargs):
+        """Create a new instance of the estimators subclass
```
adason: The docstring needs to be fixed here. Have you run documentation tool to see how this is parsed?

- input parameter should be documented (even ones that was hidden in **kwargs). 
- we should make sure to reference to `Estimator` class in the docstring, as this class is now separated from the `Estimator` class. 
- make sure the return value is clear.

sanjayuconn (commented):
Do we need prev commit history (15) for one change? Can we merge all into 1?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
+from tensorboardX import SummaryWriter
+
+import datasetinsights.constants as const
+from datasetinsights.estimators import Estimator
+from datasetinsights.storage.checkpoint import EstimatorCheckpoint
+from datasetinsights.storage.kfp_output import KubeflowPipelineWriter
+
+
+class EstimatorFactory:
```
adason: Is this class necessary? What about a module level method `create_estimator(...)` and a private module level method `_find_estimator(...)`? Does this make the high level API easier to read/understand? Which of the following pattern is better?

```
from datasetinsights.estimators import Estimator
estimator = Estimator.create(...)
```

v.s.

```
from datasetinsights.estimators import EstimatorFactory
estimator = EstimatorFactory.create(...)
```

v.s.

```
from datasetinsights.estimators import create_estimator
estimator = create_estimator(...)
```

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+            config (dict): parameters specific to each estimators subclass
+                used to create a estimators instance
+
+        Returns:
+            an instance of the specified estimators subclass
+        """
+
+        estimators_cls = EstimatorFactory.find(name)
+
+        logdir = kwargs["params"]["tb_log_dir"]
```
adason: [mustfix] These variables are required and should be explicitly unpacked by the method. Once they are unpacked, you can throw the rest to downstream class. We can set default to None if they are not provided.

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+        no_cuda = kwargs["params"]["no_cuda"]
+        data_root = kwargs["params"]["data_root"]
+        model_config = kwargs["model_config"]
+        if "kfp_metrics_dir" in kwargs:
```
adason: Use `create(,kfp_metrics_dir=const.DEFAULT_KFP_METRICS_DIR,...)` instead.

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+            kfp_metrics_dir = kwargs["params"]["kfp_metrics_dir"]
+        else:
+            kfp_metrics_dir = const.DEFAULT_KFP_METRICS_DIR
+
+        if "kfp_metrics_filename" in kwargs:
+            kfp_metrics_filename = kwargs["params"]["kfp_metrics_filename"]
+        else:
+            kfp_metrics_filename = const.DEFAULT_KFP_METRICS_FILENAME
+
+        if logdir == const.NULL_STRING:
```
adason: Please remove this hack. It's non-longer required. By default logdir will. always exist from CLI.

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+        # todo this makes it so that we lose the tensorboard
+        #  writer of non-master processes which could make debugging harder
+
+        writer = SummaryWriter(logdir,)
+        kfp_writer = KubeflowPipelineWriter(
+            filename=kfp_metrics_dir, filepath=kfp_metrics_filename,
+        )
+        checkpointer = EstimatorCheckpoint(
+            estimator_name=model_config.estimator,
+            log_dir=writer.logdir,
```
adason: Please use `checkpointdir` that was parsed from CLI directly.

datasetinsights/commands/evaluate.py
```diff
@@ -88,4 +89,15 @@ def cli(
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    # TODO: Call evaluate command here.
+
+    from datasetinsights.estimators.factory import EstimatorFactory
+
+    model_config = CN.load_cfg(open(ctx.params["config"], "r"))
```
adason: [nit] should call this `estimator_config` or `config`. Model usually means pytorch/tensorflow graph. This concpet does not apply here.

datasetinsights/commands/evaluate.py
```diff
@@ -88,4 +89,15 @@ def cli(
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    # TODO: Call evaluate command here.
+
+    from datasetinsights.estimators.factory import EstimatorFactory
```
adason: Move to top of the file.

datasetinsights/cli.py
```diff
@@ -89,22 +70,14 @@ def parse_args():
...
-    )
     parser.add_argument(
         "--data-root",
         type=str,
         default=system.data_root,
         help="root directory of all datasets",
     )
-    parser.add_argument(
+
+    parser.add_argument(  # This has to be moved to estimator config
```
adason: Is this variable moved to estimator config?

datasetinsights/commands/train.py
```diff
@@ -96,4 +97,14 @@ def cli(
...
+
+    model_config = CN.load_cfg(open(ctx.params["config"], "r"))
```
adason: Should use `config` instead of `ctx.params["config"]`.

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+            kfp_metrics_filename = kwargs["params"]["kfp_metrics_filename"]
+        else:
+            kfp_metrics_filename = const.DEFAULT_KFP_METRICS_FILENAME
+
+        if logdir == const.NULL_STRING:
+            # Use logdir=None to force using SummaryWriter default logdir,
+            # which points to ./runs/<model>_<timestamp>
+            logdir = None
+
+        # todo this makes it so that we lose the tensorboard
```
adason: What's this comments for?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+        #  writer of non-master processes which could make debugging harder
+
+        writer = SummaryWriter(logdir,)
```
adason: Why extra comma here?

datasetinsights/estimators/factory.py
sanjayuconn: Do we have unit test case of factory?

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+        Returns:
+            an instance of the specified estimators subclass
+        """
+
+        estimators_cls = EstimatorFactory.find(name)
+
+        logdir = kwargs["params"]["tb_log_dir"]
+        no_cuda = kwargs["params"]["no_cuda"]
+        data_root = kwargs["params"]["data_root"]
+        model_config = kwargs["model_config"]
```
adason: [nit] use `estimator_config`  or `config` here.

datasetinsights/commands/train.py
```diff
@@ -96,4 +97,14 @@ def cli(
...
+
+    estimator = EstimatorFactory.create(
+        name=model_config.estimator,
+        params=ctx.params,
```
adason: Should unpack params here instead of storing dict-valued parameters. I would expect this method to call `EstimatorFactory.create(name=config.etsimator, config=config, **ctx.params)`

datasetinsights/estimators/factory.py
```diff
@@ -0,0 +1,85 @@
...
+            logdir=logdir,
+            data_root=data_root,
+            no_cuda=no_cuda,
+            kfp_metrics_dir=kfp_metrics_dir,
+            kfp_metrics_filename=kfp_metrics_filename,
+        )
+
+    @staticmethod
+    def find(name):
+        """Find Estimator subclass based on the given name
```
adason: Docstring should be improved here. See previous comment.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
     ):
         """initiate estimator."""
+
         logger.info(f"initializing faster rcnn")
         self.config = config
-        self.device = device
-        self.writer = writer
-        self.kfp_writer = kfp_writer
+
+        if "RANK" in os.environ and "WORLD_SIZE" in os.environ:
```
adason, line 111: Should wrap this method (originally in init_pytorch_distributed_mode(...))in a private instance method. We will have to refactor this later, but using private instance method makes it clear these are objects that should be refactored out in __init__ method somehow.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
+
+                torch.distributed.init_process_group(
+                    backend="nccl",
+                    init_method="env://",
+                    world_size=self.world_size,
+                    rank=self.rank,
+                )
+                torch.distributed.barrier()
+
+        if torch.cuda.is_available() and not no_cuda:
```
adason: Good opportunity to use a private instance methohd.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
+            self.device = torch.device("cuda")
+        else:
+            self.device = torch.device("cpu")
+
+        self.writer = SummaryWriter(logdir, write_to_disk=self.distributed)
```
adason: [mustfix] It's important to have some document on what we discussed in the meeting. I will add a short paragraph describing this change and why we are doing this here (instead of having this taken care of in the `factory.create` method.)

Should be better using `self.writer._write_to_disk = is_master()`. (Although this tries to access class private method which is discouraged. We know this is a hack anyway...)

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
+
+        self.kfp_writer = KubeflowPipelineWriter(
```
adason: Why is this here? Isn't this already handled in `factor.created` method?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
         device: model training on device (cpu|cuda)
+        gpu (int): (optional) gpu id on which code will execute
+        rank (int): (optional) rank of process executing code
     """
 
     def __init__(
         self,
         *,
         config,
-        writer,
```
adason, line 59: Why is writer, checkpointer and kfp_writer deleted here? Should they directly take input from the instances that was initialized in the `factory.create` method?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
+        else:
+            self.device = torch.device("cpu")
+
+        self.writer = SummaryWriter(logdir, write_to_disk=self.distributed)
+
+        self.kfp_writer = KubeflowPipelineWriter(
+            filename=kfp_metrics_filename, filepath=kfp_metrics_dir,
+        )
+
+        self.checkpointer = EstimatorCheckpoint(
```
adason: I think it's better to override the instance variable directly, such as `self.checkpointer.distributed = self.distributed`

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -80,59 +141,60 @@ def __init__(
...
         self.model.to(self.device)
 
-        if self.config.system.distributed:
+        if self.distributed:
             self.model = torch.nn.parallel.DistributedDataParallel(
                 self.model, device_ids=[self.gpu]
             )
             self.model_without_ddp = self.model.module
 
+        self.checkpoint_file = self.config.get(
```
adason: [mustfix] checkpoint_file should be removed from config.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -33,44 +36,102 @@ class FasterRCNN(Estimator):
...
         config (CfgNode): estimator config
-        writer: Tensorboard writer object
-        kfp_writer: KubeflowPipelineWriter object to write logs
-        checkpointer: Model checkpointer callback to save models
-        device: model training on device (cpu|cuda)
         box_score_thresh: (optional) default threshold is 0.05
-        gpu (int): (optional) gpu id on which code will execute
-        rank (int): (optional) rank of process executing code
+        distributed: whether or not the estimator is distributed
+        data_root: Root directory on localhost where datasets are located
```
adason: Actually when the estimator is initialized, it does not care the data_root parameter as no data are loaded in this stage. I think it's better to have data_root parsed at later stage when `estimator.train(...)` is called.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -1039,3 +1101,21 @@ def __call__(self, image, target):
         """transform image to tesnor."""
         image = torchvision.transforms.functional.to_tensor(image)
         return image, target
+
+
+def is_master():
```
adason: This should be moved to `estimator.torch_distributed.py` as this actually applies to all estimators training under distributed pytorch.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -1039,3 +1101,21 @@ def __call__(self, image, target):
...
+    rank = int(os.getenv("RANK", 0))
+    return rank == 0
+
+
+def is_dist_avail_and_initialized():
```
adason: move out of this file.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -1039,3 +1101,21 @@ def __call__(self, image, target):
...
+    return dist.is_available() and dist.is_initialized()
+
+
+def get_world_size():
```
adason: move out of this file.

datasetinsights/estimators/deeplab.py
AdamPalmarUnity: I'll add back the *

kubeflow/groceries_real_train.py
```diff
@@ -24,8 +24,9 @@ def train_pipeline(
     config_file: str = (
         "datasetinsights/configs/faster_rcnn_groceries_real.yaml"
     ),
-    epochs: int = 50,
+    extra_options: str = "",
```
adason: This is not implemented. Can you remove them to reduce confusion?

kubeflow/train_pipeline.py
```diff
@@ -15,9 +15,10 @@ def train_pipeline(
     docker_image: str = (
         "gcr.io/unity-ai-thea-test/datasetinsights:<git-comit-sha>"
     ),
+    config_file: str = "datasetinsights/configs/faster_rcnn_synthetic.yaml",
     auth_token: str = "xxxxxx",
     run_execution_id: str = "EjPQYAN",
-    epochs: int = 10,
+    extra_options: str = "",
```
adason: not implemented

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: @adason Can you explain why we are removing dryrun. Also you are right the create_dryrun_dataset could be deleted if this is correct. I will wait for YC to respond

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: See above answer

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: The config.system variables have been removed from the config. The model config does not include this data.

datasetinsights/estimators/factory.py
AdamPalmarUnity: I will create the docstring

tests/test_evaluate_command.py
```diff
@@ -0,0 +1,78 @@
...
+
+import pytest
+from click.testing import CliRunner
+from yacs.config import CfgNode
+
+from datasetinsights.commands.evaluate import cli
+from datasetinsights.estimators.factory import EstimatorFactory
+
+
+@pytest.mark.parametrize(
```
adason, line 10: [nit] I found it in-general harder to understand tests when using `@pytest.mark.parametrize`. This is a double edged sword. I will recommend not to use it unless you have 3+ different parameters.

datasetinsights/estimators/factory.py
AdamPalmarUnity: Currently we do not have any unit tests for the Factory i will add some

tests/test_evaluate_command.py
```diff
@@ -0,0 +1,78 @@
...
+    # assert
+    open_mock.assert_called_once_with(
+        "tests/configs/faster_rcnn_groceries_real_test.yaml", "r"
+    )
+    cfg_node_mock.assert_called_once()
+    estimator_factory_create_mock.assert_called_once()
+    estimator_factory_create_mock.return_value.evaluate.assert_called_once()
+
+
+@pytest.mark.parametrize(
```
adason, line 52: Same as above. Why use `@pytest.mark.parametrize` for a single test input value?

adason (changes requested):

tests/test_faster_rcnn.py
AdamPalmarUnity: This directory should being created by tempfile.TemporaryDirectory() i can replace the string with a constant.

datasetinsights/commands/train.py
AdamPalmarUnity: What is the advantage of having it at the top of the file as opposed to in the function?

datasetinsights/estimators/deeplab.py
AdamPalmarUnity: I don't see any difference between the two, but i can change it if its an issue

datasetinsights/estimators/factory.py
AdamPalmarUnity: Sure will put them in the same file.

datasetinsights/estimators/factory.py
AdamPalmarUnity: Will do

datasetinsights/estimators/factory.py
AdamPalmarUnity: I would go with 

```
from datasetinsights.estimators import EstimatorFactory
estimator = EstimatorFactory.create(...)

or

from datasetinsights.estimators import create_estimator
estimator = create_estimator(...)
```
I think coupling the Estimator.create to creation of all other estimators is a violation of the single responsibility principle .
Besides that i think its a matter of preference whether its a function or a static method. I would prefer to instantiate the Factory and then call the create method but that might just be me.

kubeflow/groceries_real_train.py
AdamPalmarUnity: I can remove this although i think its a large benefit to support it at some point

kubeflow/train_pipeline.py
AdamPalmarUnity: will do

tests/test_evaluate_command.py
AdamPalmarUnity: There are two inputs. It would just be a duplication of the below code. Plus the fact that more can be added this is just the start

tests/test_evaluate_command.py
AdamPalmarUnity: The idea is also that more can be added. What do you think makes that test hard to understand besides knowing what @pytest.mark.parametrize does?

datasetinsights/estimators/factory.py
AdamPalmarUnity: Its an old comment which i left in. I think it relates to write_to_disk=is_master()

datasetinsights/estimators/factory.py
AdamPalmarUnity: This is added by the formatting tool

datasetinsights/estimators/factory.py
AdamPalmarUnity: Im not sure what you mean by explicitly unpacking them in the method. Can you elaborate what you mean by this?

datasetinsights/estimators/factory.py
AdamPalmarUnity: Will do

datasetinsights/estimators/factory.py
AdamPalmarUnity: Will do

datasetinsights/estimators/factory.py
AdamPalmarUnity: Ok

datasetinsights/commands/evaluate.py
AdamPalmarUnity: Ok

datasetinsights/commands/evaluate.py
AdamPalmarUnity: Will do

datasetinsights/commands/train.py
AdamPalmarUnity: Will do

datasetinsights/estimators/faster_rcnn.py
adason: We should be able to use DummyDetecction2D dataset for the same purpose of dry-run. The original implementation still require user to download the whole dataset in order to run train/evaluate. This does not help development process.  We will also not supply `--dry-run` in cli.

datasetinsights/commands/train.py
adason: https://www.python.org/dev/peps/pep-0008/#imports

>Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: I agree i just did not want to keep refactoring.

datasetinsights/estimators/deeplab.py
adason: In theory it will save some "lookup" time as `self.xxx` requires reference to the class object. This is very minor so it's not very important.... that's why I put a `[nit]` prefix here.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: They are deleted because they are no longer needed in the constructor as they are re-initialized later in the constructor. I dont think this is a good idea but its what we agree on during the meeting on 7 of August

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: I have keep it as close to the implementation of the reason that we do not have any tests that cover this refactoring change. Therefore i have not refactored it into any private method as requested on 7 of August

datasetinsights/estimators/factory.py
adason: What do you mean by `violation of the single responsibility principle`? 

I agree `EstimatorFactory.create(...)` and `create_estimator(...)` are better than `Estimator.create(...)`. It just seems `EstimatorFactory` over emphasized on class over simple module/method which is usually encouraged in python.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: This was specifically what was requested at the meeting from what i recall

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: For now it gets the same treatment as the other writers. It can and should be changed i think. but i did it to be consistent with the above and below writers

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: This is under the assumption that the checkpointer has been initialized with is not the case as we discussed

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: Will do

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: I can move it, but this was done to specifically encapsulate the dependency in this module.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: See above comment

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: See above comment

datasetinsights/cli.py
AdamPalmarUnity: No not yet good catch.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: WIll delete the function create_dryrun_dataset

datasetinsights/estimators/factory.py
adason: It should be something like this
```
def create_estimator(name, config, *, tb_log_dir=None, no_cuda=False, checkpointdir=None,..., **kwargs):
    writer = SummaryWriter(tb_log_dir)
    chickpointer = Chickpointer(chickpointdir, ...)
    estimator_cls = find_estimator(name)
    
    return estimator_cls(writer, checkpointer, ..., **kwargs)
    
estimator = create_estimator(config.estimantor, **ctx.params)
```
Item18-21 in this [book](https://www.oreilly.com/library/view/effective-python-59/9780134034416/) covered this topic in great detail. Highly recommend it.

datasetinsights/commands/train.py
AdamPalmarUnity: The cli will take about 5 second to respond to --help with this change. It become very slow to use and experiment with if this change is made.

datasetinsights/estimators/deeplab.py
AdamPalmarUnity: Will change it

datasetinsights/commands/train.py
adason: This rule should **NEVER** be broken except cases when the dependency is optional. I am surprised it passed `flake8` without error or some warnings. We should figure out why it's vert slow to start CLI (usually means some costly global variable exists that are computed when cli starts)

datasetinsights/estimators/factory.py
AdamPalmarUnity: The violation is having the Estimator class be responsible for multiple functions. I will change it to a function if this is preferred.

datasetinsights/estimators/factory.py
AdamPalmarUnity: Makes sense will make the change

datasetinsights/commands/train.py
AdamPalmarUnity: I think this is due to the importing of all the modules that comes with EstimatorFactory.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: Yep

datasetinsights/commands/train.py
AdamPalmarUnity: Will do

tests/test_evaluate_command.py
adason: This is exactly what makes `@pytest.mark.parametrize` harder to understand. If the code has changed slightly, this parameterization has to be completely removed, and this test has to be broken down to a few different functions. Usually you want to have 1 input that test 1 behavior, and the other input that test slightly different behavior. So within a test function, you have exactly 1 input, a few asserts that validate the outcome of that one input.

In unittests it's usually not a big concern for some code duplication. You prefer unit of isolation (hence the name unittest)

tests/test_evaluate_command.py
adason: Why have two inputs that test exactly the same behavior?

datasetinsights/estimators/factory.py
adason: A class might have multiple functions (by nature). I think either `EstimatorFactory` class or `create_estimator` function works well in this case, and they both provide good interface for users. It just feels a little weird (not pythonic) having a class with only two static methods, which makes me think these two methods might not need the encapsulation that was provided by class.

tests/test_evaluate_command.py
AdamPalmarUnity: If the parameterization changes this would mean that there is a potential break in the cli interface. 

"Usually you want to have 1 input that test 1 behavior" This is exactly why the parametrize decorator is used. They are invoked as different test just with different arguments. Which is the purpose of @pytest.mark.parametrize

tests/test_evaluate_command.py
AdamPalmarUnity: The purpose of the test is to validate that the config file is not loaded and the estimator factory is not called. 

`download evaluate
`
`download evaluate --config tests/configs/faster_rcnn_groceries_real_test.yaml --data-root invalid-data-root`


The purpose of unittests is to validate that for a given input a given output is expected. It is possible to have a program with different inputs and the same behavior. How else would you test that the behavior is the same? 


From https://en.wikipedia.org/wiki/Unit_testing
Writing and maintaining unit tests can be made faster by using parameterized tests. These allow the execution of one test multiple times with different input sets, thus reducing test code duplication. Unlike traditional unit tests, which are usually closed methods and test invariant conditions, parameterized tests take any set of parameters.

AdamPalmarUnity: From a design perspective it might make more sense to have the cli throw an exception, but this does not change the purpose of using parametrized test.

datasetinsights/cli.py
AdamPalmarUnity: I can't find what the default value is supposed to be.

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: WIll change it to a private method

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: I dont think it is a good idea to access the private variable after it has been instantiated as we don't know if it changes the behaviour of the class. Unless you are confident it does not impact the functionality.

datasetinsights/configs/faster_rcnn_groceries_real.yaml
```diff
@@ -44,4 +44,3 @@ optimizer:
 pretrained: False
 pretrained_backbone: True
 synchronize_metrics: True
-checkpoint_file: None
```
BlairLee, line 45: I believe we need this line in all YAML files. Because if it is removed, it will get error message when we do the fine-tuning. Have you checked it?

AdamPalmarUnity: @adason just to check this was what you intended right? To have it removed.

AdamPalmarUnity: Don't we have test that validates finetuning? @BlairLee

datasetinsights/estimators/faster_rcnn.py
adason: You are right. It's not a good idea to change private variable in this case. 
https://github.com/lanpa/tensorboardX/blob/34d1616c035faaa0f3f7c9d19cb8bb4425f19939/tensorboardX/writer.py#L275

datasetinsights/estimators/faster_rcnn.py
adason: `KubeflowPipelineWriter ` is not pytorch specific so I think it's safe to use the one from `factor.created` method.

datasetinsights/configs/faster_rcnn_groceries_real.yaml
adason: This has to be removed ad they don't belong to the yaml file. We should fix estimator code if it breaks something.

datasetinsights/estimators/faster_rcnn.py
adason: This is actually not what we discussed in the meeting. `factory.create` will initialized necessary object with defaults. We  have to do override just to "make it working for torch.distributed". So I will expect to see something like this

```
def create_estimator()
    writer = SummaryWriter(...) 
    kfp_writer = KubeflowPipelineWriter(...)
    checkpointer = EstimatorCheckpoint(...)

    return estimators_cls(
        writer=writer,
        kfp_writer=kfp_writer,
        checkpointer=checkpointer,
        **kwargs,
    )

class FasterRCNN:

def __init__(self, ..., writer, kfp_writer, checkpointer, ...):
    # note: this is a hack to make torch.distributed specific init in the class,.....
    self.writer = SummaryWriter(writer.logdir, writer_to_file=is_master())
    self.kfp_writer = kfp_writer

    checkpointer.distributed = is_distributed()
    self.checkpointer = checkpointer
```

In the future, we might have something like

```
class FasterRCNN:

def __init__(self, ..., writer, kfp_writer, checkpointer, ...):
    self.writer = make_it_work_with_torch_distirbuted(writer)
    self.kfp_writer =make_it_work_with_torch_distirbuted(kfp_writer)
    self.checkpointer = make_it_work_with_torch_distirbuted(checkpointer)

# and do something else to make these default "writers" work with torch_distributed. 
```

datasetinsights/datasets/synthetic.py
```diff
@@ -170,73 +167,30 @@ def __init__(
...
             random_seed (int): random seed used for splitting dataset into
                 train and val
         """
-        dataset_directory = os.path.join(data_root, const.SYNTHETIC_SUBFOLDER)
         if (
-            os.path.isdir(dataset_directory)
-            and glob.glob(f"{dataset_directory}/**/*.png")
-            and glob.glob(f"{dataset_directory}/**/*.json")
+            os.path.isdir(data_path)
+            and glob.glob(f"{data_path}/**/*.png")
```
adason: PNG files are not required for Synthetic dataset. We could change image format to jpeg or the dataset might include only LIDAR files such as `*.pcd`

datasetinsights/datasets/synthetic.py
```diff
@@ -346,7 +302,7 @@ def download_captures_from_manifest(self, manifest_file):
         Args:
             manifest_file (str): path to a manifest file
         """
-        _download_captures(self.root, manifest_file)
+        _download_captures(self.data_path, manifest_file)
```
adason: This code path is actually dead and should be removed. `self.download` is no longer called in `__init__`.

datasetinsights/datasets/synthetic.py
AdamPalmarUnity: Can we change this in a follow up PR?

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: Will make this change

datasetinsights/datasets/synthetic.py
AdamPalmarUnity: I don't think adding things to refactor to this PR is beneficial. In this PR the code path is not dead. This should be resolved in another PR i think.

adason (dismissed):

datasetinsights/estimators/deeplab.py
```diff
@@ -126,9 +125,9 @@ def __init__(self, *, config, writer, checkpointer, device, **kwargs):
         self.lr_scheduler = lr_scheduler
 
         # load estimators from file if checkpoint_file exists
-        ckpt_file = config.checkpoint_file
+        ckpt_file = self.config.checkpoint_file
```
adason: `checkpoint_file` should come from CLI, not the config yaml.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -80,57 +92,111 @@ def __init__(
...
-        if self.config.system.distributed:
+        if self.distributed:
             self.model = torch.nn.parallel.DistributedDataParallel(
                 self.model, device_ids=[self.gpu]
             )
             self.model_without_ddp = self.model.module
 
-    def train(self, **kwargs):
-        """start training, save trained model per epoch."""
+        self.checkpoint_file = checkpoint_file
```
adason: Should we have `self.ckeckpoint_file` to be an instance variable of `FasterRCNN` class?

adason (dismissed):

datasetinsights/estimators/faster_rcnn.py
AdamPalmarUnity: Do you mean a variable instead of a instance varable?

adason: Yes. Is `self.ckeckpoint_file` used in other places except init? My guess is not.

adason (approved):

PR #26, opened 2020-08-06
Writer add_scaler bug fix

AdamPalmarUnity:
# Peer Review Information

Fixing a bug in the faster_rcnn.py that produced wrong tensorboard charts. 

# Pull Request Check List


datasetinsights/constants.py
```diff
@@ -53,3 +53,6 @@
     "allow_extra_args": True,
 }
 DEFAULT_DATASET_VERSION = "latest"
+
+SUMMARY_WRITER_MAX_QUEUE = 100
+SUMMARY_WRITER_FLUSH_SECS = 30
```
BlairLee: Do you think this is the root cause of the previous error?

BlairLee:
Can you point us the kubeflow run for the 10% fine-tuning job?

datasetinsights/constants.py
AdamPalmarUnity: I cant say for sure if the parameters are the root cause, but i was not able to replicate the error with the new parameters.

adason: Out of curiosity, do you still encounter this error without changing these default parameter? I am very interested to know the root cause of this issue and it's still not clear to me why changing these parameter fixed the issue.

AdamPalmarUnity: I will run one pipeline again with default SummaryWriter values

AdamPalmarUnity: I think we need to look into the tensorboardx source code to find the root cause

AdamPalmarUnity: My hypothesis is that the SummaryWriter is trying to flush to disk and upload to gcs, but before the upload has finished it tries to upload another file causing the previous upload close its connection. Therefore increasing the max_queue size is the most important if there are alot of writes happening in fast succession after each other. Looking the source code it does make sense that the uploader to GCS.


https://github.com/lanpa/tensorboardX/blob/34d1616c035faaa0f3f7c9d19cb8bb4425f19939/tensorboardX/record_writer.py#L114
```
class GCSRecordWriter(object):
    """Writes tensorboard protocol buffer files to Google Cloud Storage."""

    def __init__(self, path):
        if not GCS_ENABLED:
            raise ImportError("`google-cloud-storage` must be installed in order to use "
                              "the 'gs://' protocol")

        self.path = path
        self.buffer = io.BytesIO()

        client = storage.Client()
        bucket_name, filepath = self.bucket_and_path()
        bucket = storage.Bucket(client, bucket_name)
        self.blob = storage.Blob(filepath, bucket)

    def __del__(self):
        self.close()

    def bucket_and_path(self):
        path = self.path
        if path.startswith("gs://"):
            path = path[len("gs://"):]
        bp = path.split("/")
        bucket = bp[0]
        path = path[1 + len(bucket):]
        return bucket, path

    def write(self, val):
        self.buffer.write(val)

    def flush(self):
        upload_buffer = copy.copy(self.buffer)
        upload_buffer.seek(0)

        self.blob.upload_from_string(upload_buffer.getvalue()) <--- My hypothesis is this is the root issue.

    def close(self):
        self.flush()
```

AdamPalmarUnity: Its based on this callback stack
```

textPayload: "Traceback (most recent call last):
  File "/usr/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/event_file_writer.py", line 220, in run
    self._record_writer.flush()
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/event_file_writer.py", line 69, in flush
    self._py_recordio_writer.flush()
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/record_writer.py", line 189, in flush
    self._writer.flush()
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/record_writer.py", line 151, in flush
    self.blob.upload_from_string(upload_buffer.getvalue())
  File "/usr/local/lib/python3.7/dist-packages/google/cloud/storage/blob.py", line 1961, in upload_from_string
    if_metageneration_not_match=if_metageneration_not_match,
  File "/usr/local/lib/python3.7/dist-packages/google/cloud/storage/blob.py", line 1795, in upload_from_file
    if_metageneration_not_match,
  File "/usr/local/lib/python3.7/dist-packages/google/cloud/storage/blob.py", line 1654, in _do_upload
    if_metageneration_not_match,
  File "/usr/local/lib/python3.7/dist-packages/google/cloud/storage/blob.py", line 1321, in _do_multipart_upload
    response = upload.transmit(transport, data, object_metadata, content_type)
  File "/usr/local/lib/python3.7/dist-packages/google/resumable_media/requests/upload.py", line 106, in transmit
    retry_strategy=self._retry_strategy,
  File "/usr/local/lib/python3.7/dist-packages/google/resumable_media/requests/_helpers.py", line 136, in http_request
    return _helpers.wait_and_retry(func, RequestsMixin._get_status_code, retry_strategy)
  File "/usr/local/lib/python3.7/dist-packages/google/resumable_media/_helpers.py", line 150, in wait_and_retry
    response = func()
  File "/usr/local/lib/python3.7/dist-packages/google/auth/transport/requests.py", line 450, in request
    **kwargs
  File "/usr/local/lib/python3.7/dist-packages/requests/sessions.py", line 530, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.7/dist-packages/requests/sessions.py", line 643, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/requests/adapters.py", line 498, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', BrokenPipeError(32, 'Broken pipe'))
" 
```

AdamPalmarUnity: Its also possible that its was a unstable connection

AdamPalmarUnity: After testing writing with the SummaryWriter locally i don't think the above hypothesis is the root cause.

BlairLee: I also think it's because of the connection since I didn't run into the issue with default settings locally.

BlairLee (dismissed):

datasetinsights/constants.py
adason: @AdamPalmarUnity @BlairLee 

So, the question still remains. Will users ran into this issue using Kubeflow pipeline? If yes, would the proposed change to increase queue size fix it? If no, we can close this PR. If this re-occurs again, we should create a new Github issue.

BlairLee: Clone two runs on kubeflow. If they have no issues, I think we are good to go.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -242,7 +242,9 @@ def train_one_epoch(
                     f"{intermediate_loss}"
                 )
                 self.writer.add_scalar(
-                    "training/running_loss", intermediate_loss, examples_seen
+                    "training/intermediate_loss",
```
adason, line 245: Also confused about what you mean by this fixed a bug. This only change the title on tensorboard from `running_loss` to `intermediate_loss`. In this case `intermediate_loss` is a better name.

BlairLee: Previously, we have two different losses written into one place, which is "training/running_loss", and the running_loss on tensorboard looks like this:
Screenshot from 2020-08-05 19-25-54

BlairLee (approved):

adason:
@AdamPalmarUnity Since the scope of this review have been changed, Can you also update the title/description before you to make sure it produces a commit message in master with good title/description?

PR #27, opened 2020-08-06
directory restructured and fixed dependency

sanjayuconn:
# Peer Review Information

- This commit includes restructured directory
- Fixed all dependent imports

# Pull Request Check List


datasetinsights/datasets/simulation/__init__.py
```diff
@@ -0,0 +1,26 @@
+from .captures import Captures
+from .download import (
+    Downloader,
+    _filter_unsuccessful_attempts,
```
adason: Since this method is private, it's probably not a good idea to be imported here.

datasetinsights/datasets/simulation/__init__.py
```diff
@@ -0,0 +1,26 @@
+from .captures import Captures
```
adason: Since we are only changing folder structure in this review, maybe this change should be pushed for the next review where you are reorganizing modules?

datasetinsights/datasets/synthetic.py
```diff
@@ -9,12 +9,13 @@
 
 import datasetinsights.constants as const
 from datasetinsights.data.bbox import BBox2D
-from datasetinsights.data.simulation import AnnotationDefinitions, Captures
-from datasetinsights.data.simulation.download import (
+from datasetinsights.datasets.simulation import (
```
adason: Based on the design doc, are we keeping the `simulation` folder? Should this be renamed to `datasetinsights.datasets.unity_perception`?

datasetinsights/stats/statistics.py
```diff
@@ -2,9 +2,12 @@
 from pathlib import Path
 
 import datasetinsights.constants as const
-from datasetinsights.data.simulation import MetricDefinitions, Metrics
-from datasetinsights.data.simulation.download import Downloader
-from datasetinsights.data.simulation.tables import SCHEMA_VERSION
+from datasetinsights.datasets.simulation import (
```
adason: datasetinsights.datasets.unity_perception?

pyproject.toml
```diff
@@ -82,7 +82,7 @@ exclude = '''
     | buck-out
     | build
     | dist
-    | datasetinsights/data/datasets/protos
+    | datasetinsights/datasets/protos
```
adason: Can you look into using general regex pattern to ignore protobuf file? How do we make sure we don't have to manually add new protobuf files to be ignored in the futrue?

adason (changes requested):

datasetinsights/datasets/simulation/__init__.py
sanjayuconn: its being used other part of the project

datasetinsights/datasets/simulation/__init__.py
sanjayuconn: Its binding in init at same level. Need to import to fix the dependency.

BlairLee (dismissed):
After you solved all the requested changes, you can merge it.

pyproject.toml
sanjayuconn: Fixed!

datasetinsights/datasets/simulation/__init__.py
adason: I didn't see it was used (except for testing) in the current master.

datasetinsights/datasets/simulation/__init__.py
adason: If this review is focused on folder structure change, these import "shortcuts" are our of the scope for this review?

datasetinsights/stats/statistics.py
sanjayuconn: Will go in next PR.

datasetinsights/datasets/synthetic.py
sanjayuconn: Will go in next PR.

pyproject.toml
BlairLee: I think what YC proposes is to use a regular expression, in case we have protobuf files in another folder.

datasetinsights/datasets/simulation/__init__.py
BlairLee: I only see this method used in testing.

adason (approved):

PR #30, opened 2020-08-10
Download cli command and UnitySimulationDownloader

AdamPalmarUnity:
# Peer Review Information

Adding download command cli changes together with the UnitySimulationDownloader.

# Pull Request Check List


datasetinsights/commands/download.py
```diff
@@ -1,13 +1,51 @@
...
+from datasetinsights.datasets.base import DatasetRegistry
+from datasetinsights.datasets.simulation.download import (
+    Downloader,
+    download_manifest,
+)
 
 logger = logging.getLogger(__name__)
 
 
+class Mutex(click.Option):
```
BlairLee: Add docstrings for this class.

datasetinsights/commands/download.py
```diff
@@ -88,43 +129,65 @@ def convert(self, value, param, ctx):
...
-    data_root,
-    source_uri,
-    include_binary,
-    dataset_version,
-    project_id,
-    run_execution_id,
-    auth_token,
+    name, source_uri, data_root, include_binary, dataset_version,
 ):
+    if not name and not source_uri:
```
BlairLee: According to error message, should it be `if not name or not source_uri`?

datasetinsights/commands/download.py
```diff
@@ -88,43 +129,65 @@ def convert(self, value, param, ctx):
...
+            source_uri=source_uri,
+            data_root=data_root,
+            include_binary=include_binary,
+        )
+
+    if source_uri.startswith("https://") or source_uri.startswith("http://"):
+        pass
+
+    if source_uri.startswith("gs://"):
+        pass
```
BlairLee: Do we need to catch invalid `source_uri` here?

datasetinsights/datasets/base.py
```diff
@@ -1,43 +1,38 @@
...
-            a label of the specified dataset subclass
-        """
-        dataset_classes = Dataset.__subclasses__()
-        dataset_names = [d.__name__ for d in dataset_classes]
-        if name in dataset_names:
-            dataset_cls = dataset_classes[dataset_names.index(name)]
-            return dataset_cls
-        else:
-            raise NotImplementedError(f"Unknown Dataset class {name}!")
+    name = ""
```
BlairLee: Why do we need this?

tests/test_faster_rcnn.py
```diff
@@ -104,8 +99,8 @@ def test_faster_rcnn_train_all(mock_create, mock_loss, config, dataset):
         kfp_writer=kfp_writer,
     )
     checkpointer.save = MagicMock()
-    train_dataset = create_dataset(config, TRAIN)
-    val_dataset = create_dataset(config, VAL)
+    train_dataset = dataset
+    val_dataset = dataset
```
BlairLee, line 113: Shall we test using both train and val dataset?

datasetinsights/commands/download.py
AdamPalmarUnity: Will do

datasetinsights/commands/download.py
AdamPalmarUnity: This issue is that this case only occurs if the the following command was called 'datasetinsights download' without --name or --source-uri so both of them are missing but they are mutually exclusive

datasetinsights/commands/download.py
AdamPalmarUnity: The invalid source_uri should be one in this class i think. Which is in the same module

```
class SourceURI(click.ParamType):
    """Source URI Parameter.
@@ -25,48 +63,51 @@
    PREFIX_PATTERN = r"^gs://|^http(s)?://|^usim://"
    def convert(self, value, param, ctx):
        """ Validate source URI and Converts the value.
        """
        match = re.search(self.PREFIX_PATTERN, value)
        if not match:
            message = (
                f"The source uri {value} is not supported. "
                f"Pattern: {self.PREFIX_PATTERN}"
            )
            self.fail(message, param, ctx)
        return value
```

tests/test_faster_rcnn.py
AdamPalmarUnity: I'm not quite sure what you mean. Functionally the test has not changed. The mocking would return the same dataset twice. Can you elaborate what you mean?

datasetinsights/datasets/base.py
AdamPalmarUnity: This is used to register datasets in the  DatasetRegistry metaclass. Since the class Dataset is a abstract base class we don't want to register it in the DatasetRegistry and have it be callable in the cli.

```
class DatasetRegistry(ABCMeta):
    registry = {}

    def __new__(cls, name, bases, namespace):

        if "name" not in namespace:
            raise RuntimeError(
                f"{name} must define a class-level dataset name!"
            )
        new_cls = super().__new__(cls, name, bases, namespace)
        if namespace["name"] != "":
            DatasetRegistry.registry[namespace["name"]] = new_cls
        return new_cls

```

tests/test_faster_rcnn.py
BlairLee: In previous version, we set train_dataset and val_dataset into different datasets. Can you set `train_dataset = dataset_train` and `val_dataset = dataset_val`?

datasetinsights/commands/download.py
BlairLee: Okay. Sounds good.

datasetinsights/datasets/base.py
```diff
@@ -1,43 +1,75 @@
...
 
-    @staticmethod
-    def create(name, **kwargs):
-        """Create a new instance of the dataset subclass
+    dataset_registry = {}
+    download_registry = {}
+    """ Internal registry for available Datasets """
 
+    @classmethod
+    def register(cls, name: str, downloaders: list):
```
BlairLee: Why do we need a list of downloaders?

datasetinsights/commands/download.py
```diff
@@ -4,10 +4,43 @@
 import click
 
 import datasetinsights.constants as const
+from datasetinsights.datasets.base import DownloaderRegistry
+from datasetinsights.datasets.unity_simulation import UnitySimulationDownloader
 
 logger = logging.getLogger(__name__)
 
 
+# class Mutex(click.Option):
```
BlairLee: So we don't use this class anymore?

tests/test_faster_rcnn.py
AdamPalmarUnity: I dont think this is the case due to the mocking of the @patch("datasetinsights.estimators.faster_rcnn.Dataset.create") This static method is called in create_dataset(config,TRAIN) and create_dataset(config,VAL). The mock makes sure that the Dataset.create static method only returns dataset fixture given in the argument dataset.

datasetinsights/commands/download.py
AdamPalmarUnity: I have commented it out for now in this draft PR. The final PR will either have it or not, but its currently unclear as the design of the CLI is not finished.

datasetinsights/datasets/base.py
AdamPalmarUnity: This is because a dataset can be downloaded in more then one way. example gs:// http:// or usim://

datasetinsights/commands/download.py
AdamPalmarUnity: Might not be used so will wait to add docstring

tests/test_faster_rcnn.py
BlairLee: Make sense.

BlairLee:
Please solve linting issues. I'm okay with the changes. If other reviewers approve it, you can merge this PR.

datasetinsights/datasets/dummy/dummy_object_detection.py
```diff
@@ -18,6 +18,8 @@ class DummyDetection2D(Dataset):
 
     """
 
+    name = ""
```
sanjayuconn: why we have name here?

datasetinsights/datasets/groceries_real.py
```diff
@@ -176,92 +174,6 @@ def _filepath(self, filename):
...
-            logger.info("The dataset file exists. Skip download.")
-            try:
-                validate_checksum(dest_path, expected_checksum)
-            except ChecksumError:
-                logger.info(
-                    "The checksum of the previous dataset mismatches. "
-                    "Delete the previously downloaded dataset."
-                )
-                os.remove(dest_path)
-        if not os.path.exists(dest_path):
```
sanjayuconn: Are we moving this under HTTPDownloader?

.flake8
```diff
@@ -1,5 +1,5 @@
 [flake8]
-max-line-length = 80
```
adason: Please stick to max-line-length 80 as described in the contribution guideline.

datasetinsights/datasets/synthetic.py
```diff
@@ -347,72 +345,3 @@ def download_captures_from_manifest(self, manifest_file):
...
-    @staticmethod
-    def unzip_file(filepath, destination):
-        """Unzips a zip file to the destination and delete the zip file.
-
-        Args:
-            filepath (str): File path of the zip file.
-            destination (str): Path where to unzip contents of zipped file.
-        """
-        with zipfile.ZipFile(filepath) as file:
-            logger.info(f"Unzipping file from {filepath} to {destination}")
```
sanjayuconn: Do we have any downloader specific to synthetic?

datasetinsights/commands/download.py
```diff
@@ -79,52 +73,14 @@ def convert(self, value, param, ctx):
         "binary files."
     ),
 )
-@click.option(
```
adason, line 82: excellent! I just the term `access token` is used in this [doc](https://github.com/Unity-Technologies/Unity-Simulation-Docs/blob/master/doc/cli.md#usim-inspect-auth). We should stick with this name for the whole repo!

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,42 @@
+import re
+from abc import ABCMeta, abstractmethod
+
+
+class DownloaderRegistry(ABCMeta):
```
adason: The downloaders are a typical Factory/Strategy pattern. Can you rename them accordingly?

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,42 @@
+import re
+from abc import ABCMeta, abstractmethod
```
sanjayuconn: Can we have docstrings at class and method level?

datasetinsights/io/source/unity_simulation.py
```diff
@@ -0,0 +1,95 @@
+import logging
```
sanjayuconn: Can we have docstrings at class and method level?

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,42 @@
...
+        else:
+            raise ValueError(f"Downloader '{source_schema}' does not exist:")
+
+    @staticmethod
+    def list_datasets():
+        return DownloaderRegistry.registry.keys()
+
+
+class DatasetDownloader(metaclass=DownloaderRegistry):
+    SOURCE_SCHEMA = ""
```
adason: Rename this to `protocol` to make it align with different downloader URI protocol such as `gs://`, `usim://` and `https://`...?

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,42 @@
...
+            raise RuntimeError(
+                f"{name} must define a class-level SOURCE_SCHEMA"
+            )
+        new_cls = super().__new__(cls, name, bases, namespace)
+        if namespace["SOURCE_SCHEMA"] != "":
+            DownloaderRegistry.registry[namespace["SOURCE_SCHEMA"]] = new_cls
+        return new_cls
+
+    @classmethod
+    def find(cls, source_uri):
```
adason: How do we enable automatically register new protocols if a new downloader (such as `s3://`), is added to the repo?

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,42 @@
+import re
```
adason: Would it be better to have `datasetinsights/io/downloader` as everything in this module will be related to different downloader?

datasetinsights/datasets/groceries_real.py
```diff
@@ -176,92 +174,6 @@ def _filepath(self, filename):
         """
         return os.path.join(self.root, self.version, filename)
 
-    @staticmethod
```
adason: Why are these deleted if the scope of this review covers only UnitySimulationDownloader and the interface design?

datasetinsights/datasets/base.py
```diff
@@ -1,44 +1,10 @@
-from abc import ABCMeta, abstractmethod
+from abc import abstractmethod
 
 
-class Dataset(metaclass=ABCMeta):
+class Dataset:
     """Abstract base class for datasets.
     """
 
-    @staticmethod
```
adason: Why are these removed in this review? Why is ABCMeta of `Dataset`  removed?

adason (changes requested):

datasetinsights/io/source/unity_simulation.py
```diff
@@ -0,0 +1,95 @@
...
+import os
```
sanjayuconn: Docstring missing

tests/test_faster_rcnn.py
```diff
@@ -254,16 +244,15 @@ def test_faster_rcnn_log_metric_val(mock_create, config, dataset):
         kfp_writer=kfp_writer,
     )
     epoch = 0
-    estimator.log_metric_val(dataset.label_mappings, epoch)
+    estimator.log_metric_val({"1": "car", "2": "bike"}, epoch)
```
sanjayuconn, line 248: Why we are creating new object {"1": "car", "2": "bike"} when we have same in dataset?

tests/test_faster_rcnn.py
```diff
@@ -312,28 +303,22 @@ def test_faster_rcnn_load(mock_create, config, dataset):
...
     train_dataset = create_dryrun_dataset(config, train_dataset, TRAIN)
     assert config.train.batch_size * 2 == len(train_dataset)
 
 
-@patch("datasetinsights.estimators.faster_rcnn.Dataset.create")
-def test_create_sampler(mock_create, config, dataset):
+def test_create_sampler(config, dataset):
     """test create sampler."""
-    mock_create.return_value = dataset
-    train_dataset = create_dataset(config, TRAIN)
```
sanjayuconn, line 336: why are we creating local reference train_dataset = dataset, why can't we use dataset directly?

tests/test_faster_rcnn.py
```diff
@@ -342,12 +327,10 @@ def test_create_sampler(mock_create, config, dataset):
 
 
 @patch("datasetinsights.estimators.faster_rcnn.torch.utils.data.DataLoader")
-@patch("datasetinsights.estimators.faster_rcnn.Dataset.create")
-def test_dataloader_creator(mock_create, mock_loader, config, dataset):
+def test_dataloader_creator(mock_loader, config, dataset):
     """test create dataloader."""
-    mock_create.return_value = dataset
```
sanjayuconn, line 348: Can we use dataset directly?

tests/test_faster_rcnn.py
```diff
@@ -359,12 +342,11 @@ def test_dataloader_creator(mock_create, mock_loader, config, dataset):
 
 
 @patch("datasetinsights.estimators.faster_rcnn.torch.utils.data.DataLoader")
-@patch("datasetinsights.estimators.faster_rcnn.Dataset.create")
-def test_create_dataloader(mock_create, mock_loader, config, dataset):
+def test_create_dataloader(mock_loader, config, dataset):
     """test load data."""
-    mock_create.return_value = dataset
+
```
sanjayuconn, line 343: Can we use dataset directly?

.flake8
AdamPalmarUnity: Why is this not 100 in the contribution guideline? :(

datasetinsights/commands/download.py
AdamPalmarUnity: What do you mean by "just the term"?

datasetinsights/datasets/dummy/dummy_object_detection.py
AdamPalmarUnity: I forgot to remove it. Will remove it

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: Originally the PR was broader as the scope was all the downloaders. I can revert this module if needed although it will no longer be called anywhere from the cli.

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: What specifically are you referring to? These methods should be in the HTTPDownloader and not the Dataset class if that is what you mean.

datasetinsights/datasets/synthetic.py
AdamPalmarUnity: From my understanding synthetic should also be using the HTTPdownloader for this release. So the answer for now is no unless it turns out its possible to merge the old dataset downloader methods

datasetinsights/io/source/base.py
AdamPalmarUnity: I think it might make sense, but there is a module called downloader.py in datasetinsights/io/ already so we might need to move this or rename it

datasetinsights/io/source/base.py
AdamPalmarUnity: Will add this

datasetinsights/io/source/base.py
AdamPalmarUnity: We would need to refactor 2 different places this class we would need to add the s3 to the pattern of the re.compile method and in the SourceURI class that validate the source-uri.
The downloader is automatically registered if it is a subclass of DatasetDownloader if that is what your question is?

datasetinsights/io/source/base.py
AdamPalmarUnity: Will do

datasetinsights/io/source/unity_simulation.py
AdamPalmarUnity: Will add docsstrings

datasetinsights/io/source/unity_simulation.py
AdamPalmarUnity: Will add docstring

tests/test_faster_rcnn.py
AdamPalmarUnity: If you look at the test what is needed for the test is not the dataset but the label_mappings. Therefore i refactored the test to no longer include the dataset as an argument

tests/test_faster_rcnn.py
AdamPalmarUnity: There is no good reason, i can change this if its more readable

tests/test_faster_rcnn.py
AdamPalmarUnity: Yes i can change this.

tests/test_faster_rcnn.py
AdamPalmarUnity: Yes

datasetinsights/datasets/base.py
AdamPalmarUnity: This was due to the fact that we no longer have the concept of name for the dataset, maybe i have misunderstood something.

datasetinsights/io/source/base.py
AdamPalmarUnity: Do you want to rename the DownloaderRegistry or the new_cls variable?

datasetinsights/datasets/base.py
AdamPalmarUnity: I'll revert it

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: I'll revert this

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: I'm reverting this in this PR

tests/datasets/test_groceriesreal.py
```diff
@@ -1,107 +0,0 @@
-import os
```
AdamPalmarUnity: @adason  Should i add back in this test?

datasetinsights/io/source/base.py
Saurav-D: I also agree with @adason with having a `downloader` subdirectory in `io`. But can we name it `dataset_downloader` since it's specific to downloading datasets. Also, @AdamPalmarUnity the module is `download.py`.

datasetinsights/io/source/base.py
```diff
@@ -0,0 +1,40 @@
...
+from abc import ABCMeta, abstractmethod
+
+
+class DownloaderRegistry(ABCMeta):
+    registry = {}
+
+    def __new__(cls, name, bases, namespace):
+        protocal = "PROTOCAL"
```
Saurav-D: Is this supposed to be `PROTOCOL`? Or `PROTOCAL` means something else?

datasetinsights/io/download.py
```diff
@@ -67,6 +67,21 @@ def download_file(source_uri: str, dest_path: str, use_cache: bool = True):
...
+        algorithm (str): checksum algorithm. Defaults to CRC32
+
+    Returns:
+        True if the file checksum matches.
+    """
+    computed = compute_checksum(filepath, algorithm)
+    return computed == expected_checksum
+
+
 def validate_checksum(filepath, expected_checksum, algorithm="CRC32"):
```
Saurav-D, line 85: How will we now validate checksum since now `DatasetDownloader` is not coupled to dataset class. How would we know the type of dataset being downloaded?

datasetinsights/io/source/base.py
AdamPalmarUnity: This is a typo good catch

datasetinsights/io/source/base.py
AdamPalmarUnity: I'm fine with naming the subdirectory downloader or dataset_downloader.

datasetinsights/io/download.py
AdamPalmarUnity: Why is the type of dataset being downloader relevant for what the checksum is? Shouldn't the checksum be stored together with the dataset instead of inside the repo?

BlairLee: @adason  and I discussed about this. We decided to put the source checksums in the repo, which is easy to implement.

datasetinsights/commands/download.py
```diff
@@ -42,20 +43,6 @@ def convert(self, value, param, ctx):
...
-    help="The dataset registry name.",
-)
-@click.option(
-    "-d",
-    "--data-root",
-    type=click.Path(exists=True, file_okay=False, writable=True),
-    default=const.DEFAULT_DATA_ROOT,
-    help="Root directory on localhost where datasets should be downloaded.",
-)
 @click.option(
```
adason, line 46: I think it might be better that we make source-uri required parameter. Otherwise, the downloader usually don't know where it should download from.

datasetinsights/commands/download.py
```diff
@@ -80,51 +74,19 @@ def convert(self, value, param, ctx):
...
-@click.option(
-    "--auth-token",
-    type=click.STRING,
-    default=None,
-    help=(
-        "Unity Simulation auth token. This will override synthetic "
-        "datasets source-uri for Unity Simulation."
-    ),
+    help="Access-token to be used to authenticate"
+    " to unity simulation for downloading the dataset",
```
adason: [nit] CAPS for `Unity Simulation`. The description is a little harder to read (with three `to`). Maybe we should rephrase it a little bit.

datasetinsights/io/downloader/base.py
```diff
@@ -0,0 +1,45 @@
+import re
+from abc import ABCMeta, abstractmethod
+
+
+class DownloaderRegistry(ABCMeta):
+    """ This class registers DatasetDownloader
```
adason: Please improve this docstring to make it clear for users what the `Registry` is used.

datasetinsights/io/downloader/base.py
```diff
@@ -0,0 +1,45 @@
...
+        subclasses based on the PROTOCOL attribute
+
+    """
+
+    registry = {}
+
+    def __new__(cls, name, bases, namespace):
+        protocol = "PROTOCOL"
+        if protocol not in namespace:
```
adason: Why is interface validation not part of the `DatasetDownloader ` interface definition? Why is the factory method responsible of performing interface validation? Also, it might be better to raise `NotImplementedError` or other errors that is explicit about interface issues. `RuntimeError` was too generic. 

I think this logic can be simplified if we create an `__init_subclass__` method in `DatasetDownloader ` such as:

```
class DatasetDownloader(metaclass=ABCMeta):

@classmethod
    def __init_subclass__(cls):
        """ Executed during subclass init.

        Raises:
            NotImplementedError: Raised if the subclass does not define
                required variables.
        """
        for var in ["PROTOCAL"]:
            if not hasattr(cls, var):
                raise NotImplementedError(
                    f"Subclass needs to have a constant class variable '{var}'."
                )
```

datasetinsights/io/downloader/base.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            raise ValueError(f"Downloader '{protocol}' does not exist:")
+
+    @staticmethod
+    def list_datasets():
+        return DownloaderRegistry.registry.keys()
+
+
+class DatasetDownloader(metaclass=DownloaderRegistry):
+    PROTOCOL = ""
```
adason: Why provide default `PROTOCOL` value here?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+from pathlib import Path
+
+from datasetinsights.io.downloader.base import DatasetDownloader
+from datasetinsights.io.usim import Downloader, download_manifest
+
+logger = logging.getLogger(__name__)
+
+
+class UnitySimulationDownloader(DatasetDownloader):
+    """ Downloads unity simulation datasets
```
adason: This docstring needs to be improved. First, you should describe `Unity Simulation` or at least provide links to official documentation. Second, we should document expected source uri as well as different override behavior here. Third, all `Args` and `Attributes` should be documented.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+        subfolder = Path(manifest_file).stem
+        output_directory = os.path.join(output, subfolder)
+        dl_worker = Downloader(manifest_file, output_directory)
+        dl_worker.download_references()
+        dl_worker.download_metrics()
+        dl_worker.download_captures()
+        if include_binary:
+            dl_worker.download_binary_files()
+
+    def parse_source_uri(self, source_uri):
```
adason, line 103: Can you add docstrings for this method?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+    """
+
+    PROTOCOL = "usim://"
+
+    def __init__(self, access_token=None):
+        self.access_token = access_token
+        self.run_execution_id = None
+        self.project_id = None
+
+    def download(self, source_uri, output, include_binary):
```
adason: We should add **kwargs to capture other unwanted parameters.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+    """ Downloads unity simulation datasets
+        Args:
+        access_token: "Access token to be used to authenticate to
+         unity simulation for downloading the dataset"
+
+    """
+
+    PROTOCOL = "usim://"
+
+    def __init__(self, access_token=None):
```
adason: Although not supported to override source_uri, we should allow `run_execution_id` and `project_id` to be part of the arguments in the `__init__` method and default to `None`. Also, this should accept `**kwargs` to capture unwanted parameters.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+        self.access_token = access_token
+        self.run_execution_id = None
+        self.project_id = None
+
+    def download(self, source_uri, output, include_binary):
+        """
+
+        Args:
+            source_uri: This is the downloader-uri that indicates where on
```
adason, line 64: If source_uri, output is always required positional argument, it should be made clear in the `DatasetDownloader` interface definition.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+             unity simulation the dataset should be downloaded from.
+
+            output: This is the path to the directory
```
adason, line 74: Have you build documentation to check if this docstring is parsed properly?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+            where the download will store the dataset.
+
+            include_binary: Whether to download binary files
```
adason, line 77: [nit] fix indentation to improve readability.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+             such as images or LIDAR point
+            clouds. This flag applies to Datasets where metadata
+            (e.g. annotation json, dataset catalog, ...) can be separated from
+            binary files.
+
+        """
+        self.parse_source_uri(source_uri)
+        # use_cache = not args.no_cache
```
adason: Remove unwanted comment.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+                self.access_token,
+                project_id=self.project_id,
+            )
+        else:
+            logger.info(
+                f"No auth token is provided. Assuming you already have "
+                f"a manifest file located in {manifest_file}"
+            )
+
+        subfolder = Path(manifest_file).stem
```
adason: This has hidden assumption that `<output_dir>/<run-execution-id>` is the root directly of the dataset (for `--train-data` flag in the downstream task). We should remove this assumption and make `<output_dir>` the root directory of this dataset.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+            not self.access_token
+            and not self.project_id
+            and not self.run_execution_id
+        ):
+            raise ValueError(
+                f"{source_uri} needs to be in format"
+                f" usim://access_token@project_id/run_execution_id"
+            )
+
+    def _parse_with_no_access_token(self, source_uri):
```
adason: Why do we need different regex to parse a string that was similar? Regex is very powerful tool to help you parsing similar looking strings with different optional components. Can you write a better regex pattern so that it can parse source URI strings correctly in one pass? Consider using `(...)?` pattern.  Also, the regex pattern should be a class constant variable so that it's easier to find/modify/improve in the future.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,116 @@
...
+        match = re.compile(r"usim://(\w+-\w+-\w+-\w+-\w+)/(\w+)")
```
adason: [nit] `[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}` might be a better UUID pattern.

adason (changes requested):

datasetinsights/commands/download.py
adason: typo. `I just realized the term....` :

datasetinsights/io/source/base.py
adason: `DownloaderRegistry` implies it's a registry which allow users to find a dataset location given a name (or unique string). We already abandoned `registry` approach during the design review, and I think it should be renamed accordingly to reduce confusion.

datasetinsights/io/source/base.py
adason: Since we already know the protocol (gs://, http://...) of all subclasses, can we dynamically determine from source URI to each downloader class without making code changes?

Some suggestions:

```
protocols = [d.PRORO for d in DatasetDownloader.subclasses()]
pattern = re.compile("|".join( protocols))
# or 
protocol_to_cls = {d.PRORO: d for d in DatasetDownloader.subclasses()}
protocol = parse_source_uri_to_get_protocol(...)
downloader_cls = protocol_to_cls[protocol]
```

datasetinsights/io/download.py
adason: Checksum is not part of this review. Currently for UnitySimulationDownloader there are no way to validate checksums as they are not provided in the manifest file.

datasetinsights/io/downloader/base.py
AdamPalmarUnity: I can use `__init__subclass__` but i don't see the point of metaclass be ABCMeta. From my understanding the point of `__init__subclass__` is to avoid having to deal with metaclasses.

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: I agree with accepting **kwargs in the constructor. Ii don't think it makes sense to have run_execution_id and project_id be a part of the constructor. If anything they could be passed to the download method with default=None. The reason being it does not make sense to reinitialize the Downloader every time a new dataset has to be downloaded.

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: WIll do

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: No not yet. Before all conversations are resolved

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: Will do

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: Will change Downloader to take the output directly in the constructor. For the future it might make more sense to have the 'manifest_file' and 'output' be taken in the various download methods instead of the constuctor.

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: Will change the regex

datasetinsights/io/downloader/base.py
AdamPalmarUnity: Not needed anymore

datasetinsights/commands/download.py
```diff
@@ -80,51 +73,19 @@ def convert(self, value, param, ctx):
...
-    project_id,
-    run_execution_id,
-    auth_token,
+    source_uri, output, include_binary, access_token,
 ):
     ctx = click.get_current_context()
     logger.debug(f"Called download command with parameters: {ctx.params}")
-    # TODO: Call download dataset command here.
+
+    downloader = DownloaderFactory.find(source_uri)(access_token=access_token)
```
adason: The name interface is a little confusing here. `DownloaderFactory.find` will actually create a downloader instance, which suggests this method should be renamed. What about a `create_downloader` method in `downloader/base.py`, and a separate method `_find_downloader`, as we will have similar pattern for `Estimators`, `Dataset`, `EvaluationMetric`,...?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,120 @@
...
+            include_binary: Whether to download binary files
+                such as images or LIDAR point
+                clouds. This flag applies to Datasets where metadata
+                (e.g. annotation json, dataset catalog, ...)
+                 can be separated from binary files.
+
+        """
+        self.parse_source_uri(source_uri)
+        manifest_file = os.path.join(output, f"{self.run_execution_id}.csv")
+        if self.access_token:
```
adason: If `access_token` is not found or if it's not valid, Maybe it's better to raise exception?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,120 @@
...
+        Args:
+            source_uri: Parses source-uri in the following format
+            usim://access-token@project-id/run-execution-id
+            or
+            usim://project-id/run-execution-id
+
+        """
+        pattern = re.compile(self.SOURCE_URI_PATTERN)
+        if pattern.findall(source_uri):
+            (access_token, project_id, run_execution_id,) = pattern.findall(
```
adason, line 116: What if `pattern.findall` found multiple matches? Will it be better if we raise exceptions?

adason (changes requested):

tests/datasets/test_groceriesreal.py
adason: Yes. The changes in this PR only focus on `DatasetDownloader` interface and `UnitySimulationDownloader` and does not make changes to groceriesreal dataset.

datasetinsights/io/downloader/unity_simulation.py
adason: In this case, I think it makes sense not to have instance variable for `run-execution-id` and `project_id`. These two parameters should be input of `UnitySimulationDatasetDownloader.download` method. When the Downloader is initialized it only have required info such as connection (e.g. GCS client)  and auth info (e.g. access token, service account,...) 

This will greatly simplify the "override" logic when `download command is called.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Args:
+        access_token: "Access token to be used to authenticate to
+         unity simulation for downloading the dataset"
+
+    """
+
+    SOURCE_URI_PATTERN = (
+        r"usim://([^@]*)?@?"
+        r"([a-fA-F0-9]{8}-"
```
adason: [nit] might be OK to have UUID regex in a single line (improve readability). If this violates pep8, you can add `# noqa: E501` to skip flake8 error for this line.

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,120 @@
...
+        r"[a-fA-F0-9]{4}-"
+        r"[a-fA-F0-9]{12})"
+        r"/(\w+)"
+    )
+
+    def __init__(self, access_token=None, **kwargs):
+        """
+
+        Args:
+            access_token: "Access token to be used to authenticate to
```
adason: why extra quote here?

datasetinsights/io/downloader/unity_simulation.py
```diff
@@ -0,0 +1,120 @@
+import logging
+import os
+import re
+
+from datasetinsights.io.downloader.base import DatasetDownloader
+from datasetinsights.io.usim import Downloader, download_manifest
```
adason: How is `io.usim.Downloader` different from `UnitySimulationDownloader`? Should we rename module or move file around to make it easier to understand? Thoughts?

datasetinsights/commands/download.py
AdamPalmarUnity: DownloaderFactory.find does not create an instance of the downloader it only returns the class. the construction happens in this line. (access_token=access_token)

AdamPalmarUnity: I'll convert it into a function instead of a static method

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: I would move the code into the UnitySimulationDownloader. As its not different in the usecase

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: will do

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: Not needed

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: This is not part of this review. its directly taken from the previous implementation
   logger.info(
                f"No auth token is provided. Assuming you already have "
                f"a manifest file located in {manifest_file}"
            )

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: I don't understand what you mean. Could you elaborate a bit?

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: This will require even more refactoring as the Downloader class is used in synthtic.py and statistics.py

datasetinsights/io/downloader/unity_simulation.py
adason: If users has input `http://domain.file usim://uuid/run` as a single string, it will actually silently fail to download the second one. This in fact should not be a valid case as our implementation should support only single source uri.

datasetinsights/io/downloader/base.py
```diff
@@ -0,0 +1,52 @@
...
+    else:
+        raise ValueError(f"Downloader not found for source-uri '{source_uri}'")
+
+    if protocol.startswith(("https://", "http://")):
+        protocol = "http://"
+
+    return _registry.get(protocol)
+
+
+def create_downloader(source_uri, **kwargs):
```
adason, line 34: This is a public method which should have docstrings. Might be better moving part of docstring in `_find_downloader` here?

datasetinsights/io/downloader/unity_simulation.py
adason: I don't agree this is not part of the review. This seems to be an opportunity to fix "cache", or think about how we want to handle such cases. I am simply pointing out in the desired behavior, what should we do "If access_token is not found or if it's not valid". What's the desired behavior? What information is reverent and you want user to know? What should pass/fail?

datasetinsights/io/downloader/unity_simulation.py
adason: Both ` synthtic.py` and `statistics.py` should not have logic to download dataset. They should be simply loading data (assuming dataset exists in local folder)

AdamPalmarUnity: I know they should not have the logic but they do currently. I'm just trying to point out that this increases the scope of the PR.

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: This code was already checked into master i just moved it. So it should have been review before

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: Are you thinking of the scenario when a user imports this class and calls parse_source_uri(source_uri) directly? The example you gave above should be stopped by the cli.

AdamPalmarUnity: What is the input being given to?

datasetinsights/io/downloader/unity_simulation.py
AdamPalmarUnity: This was the original code in the usim_download.py script 
```

def run(args):
    if args.verbose:
        root_logger = logging.getLogger()
        root_logger.setLevel(logging.DEBUG)

    logger.info("Run command with args: %s\n", args)
    data_root = args.data_root

    auth_token = args.auth_token
    use_cache = not args.no_cache
    run_execution_id = args.run_execution_id

    manifest_file = os.path.join(
        data_root, const.SYNTHETIC_SUBFOLDER, f"{run_execution_id}.csv"
    )
    if auth_token:
        manifest_file = download_manifest(
            run_execution_id,
            manifest_file,
            auth_token,
            project_id=args.project_id,
            use_cache=use_cache,
        )
    else:
        logger.info(
            f"No auth token is provided. Assuming you already have "
            f"a manifest file located in {manifest_file}"
        )

    subfolder = Path(manifest_file).stem
    root = os.path.join(data_root, const.SYNTHETIC_SUBFOLDER, subfolder)

    dl_worker = Downloader(manifest_file, root, use_cache=use_cache)
    dl_worker.download_references()
    dl_worker.download_metrics()
    dl_worker.download_captures()
    if args.include_binary:
        dl_worker.download_binary_files()
```

adason (approved):

PR #32, opened 2020-08-12
Improved Testing Time

Saurav-D:
# Peer Review Information

- Fix test, and mocked up methods for improving testing time from about 500s to 90s.
- Removed `test_clean_dir` which generated warning because it was trying to delete temp dir which automatically gets deleted after the tests finish.

# Pull Request Check List

### DIFF
--- a/tests/test_faster_rcnn.py
+++ b/tests/test_faster_rcnn.py
@@ -86,16 +86,27 @@ def test_faster_rcnn_train_one_epoch(mock_create, config, dataset):
     )
 
 
+@patch("datasetinsights.estimators.faster_rcnn.FasterRCNN.train_one_epoch")
 @patch("datasetinsights.estimators.faster_rcnn.Loss.compute")
 @patch("datasetinsights.estimators.faster_rcnn.Dataset.create")
-def test_faster_rcnn_train_all(mock_create, mock_loss, config, dataset):
+def test_faster_rcnn_train_all(
+    mock_create, mock_loss, mock_train_one_epoch, config, dataset
+):
     """test train on all epochs."""
     loss_val = 0.1
     mock_create.return_value = dataset
     mock_loss.return_value = loss_val
+    log_dir = tmp_name + "/train/"
+    config.system.logdir = log_dir
     writer = MagicMock()
     kfp_writer = MagicMock()
-    checkpointer = MagicMock()
+
+    checkpointer = EstimatorCheckpoint(
+        estimator_name=config.estimator,
+        log_dir=log_dir,
+        distributed=config.system["distributed"],
+    )
+
     estimator = FasterRCNN(
         config=config,
         writer=writer,
@@ -103,7 +114,6 @@ def test_faster_rcnn_train_all(mock_create, mock_loss, config, dataset):
         checkpointer=checkpointer,
         kfp_writer=kfp_writer,
     )
-    checkpointer.save = MagicMock()
     train_dataset = create_dataset(config, TRAIN)
     val_dataset = create_dataset(config, VAL)
     label_mappings = train_dataset.label_mappings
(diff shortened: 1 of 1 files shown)


tests/test_faster_rcnn.py
```diff
@@ -86,24 +85,34 @@ def test_faster_rcnn_train_one_epoch(mock_create, config, dataset):
...
     loss_val = 0.1
     mock_create.return_value = dataset
     mock_loss.return_value = loss_val
+    log_dir = tmp_name + "/train/"
+    config.system.logdir = log_dir
     writer = MagicMock()
     kfp_writer = MagicMock()
-    checkpointer = MagicMock()
+
+    checkpointer = EstimatorCheckpoint(
```
BlairLee, line 103: Why do we need to remove checkpointer mocking here?

tests/test_faster_rcnn.py
```diff
@@ -426,8 +434,3 @@ def test_faster_rcnn_predict(mock_create, config, dataset):
     image = torchvision.transforms.functional.to_tensor(image)
     result = estimator.predict(image)
     assert result == []
-
-
-def test_clean_dir():
```
BlairLee, line 431: If we delete this, will the temp folder be there forever?

tests/test_faster_rcnn.py
Saurav-D: Because we are mocking `train_loop` in `test_faster_rcnn_train` and it'll not save the checkpointer which is needed in later tests.

tests/test_faster_rcnn.py
Saurav-D: temp folder gets deleted automatically after the tests execution

BlairLee: I remember Sanjay talked about if we don't remove it, it would be still in memory. @sanjayuconn , can you elaborate on this?

Saurav-D:
Screen Shot 2020-08-03 at 11 19 46 PM
if we keep it then we get the following warning, which says its trying to delete the temp dir but its not present.

tests/test_faster_rcnn.py
Saurav-D: Screen Shot 2020-08-03 at 11 19 46 PM
if we keep it then we get the following warning, which says its trying to delete the temp dir but its not present. @BlairLee

sanjayuconn:
Changes looks good to me. If Bowen is fine, you can merge.

sanjayuconn (commented):
Changes looks good to me. If Bowen is fine, you can merge.

BlairLee (dismissed):

tests/test_faster_rcnn.py
sanjayuconn: https://docs.python.org/3.4/library/tempfile.html#tempfile.TemporaryDirectory

- We are creating: 
tmp_dir = tempfile.TemporaryDirectory()

- so it requires manual clean up.

if we create using below
with TemporaryDirectory() as temp_dir:
    # ... do something with temp_dir
# automatically cleaned up when context exited

tests/test_faster_rcnn.py
sanjayuconn: https://docs.python.org/3.4/library/tempfile.html#tempfile.TemporaryDirectory

- We are creating: 
tmp_dir = tempfile.TemporaryDirectory()

- so it requires manual clean up according to doc. But in our case its cleaning up automatically so  we may want to check before calling clean method.

if we create using below
with TemporaryDirectory() as temp_dir:
    # ... do something with temp_dir
#.... automatically cleaned up when context exited

Saurav-D: @sanjayuconn according to tests the checkpoint file created in `train_all_test` is required by `predict` and `load` tests. If we use context manager they'll not be able to use it. But what we can do is have a check before we try to remove temp dir.

sanjayuconn (approved):

PR #36, opened 2020-08-12
user import fixed and moved storage to io

sanjayuconn:
# Peer Review Information

This commit includes:
1. Transfer of storage directory's modules to io and fix all dependencies
2. User imports fixed and cleaned imports in notebook

# Pull Request Check List


datasetinsights/datasets/__init__.py
```diff
@@ -3,7 +3,7 @@
 from .coco import CocoDetection
 from .groceries_real import GoogleGroceriesReal, GroceriesReal
 from .nyudepth import NyuDepth
-from .synthetic import SynDetection2D
+from .synthetic import SynDetection2D, read_bounding_box_2d
```
BlairLee, line 6: I think `read_bounding_box_2d` method should be in `io/bbox.py`, cause it's related to 2d bounding box.

datasetinsights/estimators/__init__.py
```diff
@@ -1,6 +1,12 @@
...
+from .faster_rcnn import FasterRCNN, convert_bboxes2canonical
 
-__all__ = [Estimator, DeeplabV3, DenseDepth, FasterRCNN]
+__all__ = [
+    Estimator,
+    DeeplabV3,
+    DenseDepth,
+    FasterRCNN,
+    convert_bboxes2canonical,
+]
```
BlairLee, line 12: **all** should be a list of strings instead of a list of class names.

datasetinsights/io/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .bbox import BBox2D
+from .checkpoint import EstimatorCheckpoint
+from .kfp_output import KubeflowPipelineWriter
+
+__all__ = [
+    BBox2D,
+    EstimatorCheckpoint,
+    KubeflowPipelineWriter,
+]
```
BlairLee, line 9: all should be a list of strings instead of a list of class names.

datasetinsights/stats/__init__.py
```diff
@@ -0,0 +1,9 @@
+from .visualization.plots import (
+    bar_plot,
+    grid_plot,
+    histogram_plot,
+    plot_bboxes,
+    rotation_plot,
+)
+
+__all__ = [bar_plot, grid_plot, histogram_plot, plot_bboxes, rotation_plot]
```
BlairLee: all should be a list of strings instead of a list of class names.

datasetinsights/datasets/__init__.py
adason: I think there are arguments that can go either way. This method reads bounding box from synthetic bounding box format to `BBox2D`. So it's related to 2D object detection as well. I don't feel strong one way or the other, as long as the module is well documented.

BlairLee: Okay. Sounds good.

sanjayuconn: Synthetic is using the bbox to transform into bbox representation so I think its right place to keep.

BlairLee (dismissed):

BlairLee (approved):

PR #39, opened 2020-08-12
Pose estimation

JonathanLeban:
This is the code to perform pose estimation. I added three different models and thus their corresponding config. I also added a metric, a dataset and a pipeline for training on Kubeflow


datasetinsights/configs/vgg_slam.yaml
```diff
@@ -0,0 +1,31 @@
...
+test:
+  dataset: synthetic
+  batch_size: 4
+metric: AMSE
+optimizer:
+  name: Adam
+  args:
+    lr: 0.0001
+    beta_1: 0.9
+    beta_2: 0.999
```
BlairLee, line 22: We may want to set beta_1 and beta_2 as default values. Remove these two variables from all yaml files.

datasetinsights/datasets/single_cube.py
```diff
@@ -0,0 +1,146 @@
...
+                    for line in file_in:
+                        y_list.append(json.loads(line.split("\n")[0]))
+                y_list += y_list1
+        df_data = pd.DataFrame.from_dict(y_list, orient='columns')
+        df_data['screenCaptureName'] = df_data['screenCaptureName'].apply(
+            lambda x: x + '.png')
+
+        return df_data
+
+    def download(self):
```
BlairLee, line 128: If this is an abstract method, then decorate it using `@abstractmethod `

datasetinsights/estimators/__init__.py
```diff
@@ -2,5 +2,9 @@
 from .deeplab import DeeplabV3
 from .densedepth import DenseDepth
 from .faster_rcnn import FasterRCNN
+from .vgg_slam import VGGSlam
+from .vgg_slam_c_s import VGGSlamCS
+from .vgg_slam_multi_objects import VGGSlamMO
 
-__all__ = [Estimator, DeeplabV3, DenseDepth, FasterRCNN]
+
+__all__ = [Estimator, DeeplabV3, DenseDepth, FasterRCNN, VGGSlam, VGGSlamCS, VGGSlamMO]
```
BlairLee, line 10: **__all__** should be a list of strings instead of a list of class names.

datasetinsights/evaluation_metrics/average_mean_square_error.py
```diff
@@ -0,0 +1,50 @@
...
+    \{\frac{1}{n}\sum_{p}^{n}{(y_p-\hat{y_p})}^2}
+"""
+
+import numpy as np
+
+from .base import EvaluationMetric
+from .exceptions import NoSampleError
+
+
+class AverageMeanSquareError(EvaluationMetric):
```
BlairLee, line 15: Can you add a unittest for this? Here is an [example](https://github.com/Unity-Technologies/datasetinsights/blob/master/tests/evaluation_metrics/test_average_relative_error.py).

datasetinsights/evaluation_metrics/average_quaternion_error.py
```diff
@@ -0,0 +1,50 @@
+r"""Average Quaternion error metric.
+"""
+
+import numpy as np
+
+from .base import EvaluationMetric
+from .exceptions import NoSampleError
+
+
+class AverageQuaternionError(EvaluationMetric):
```
BlairLee, line 10: Can you add a unittest for this? Here is an [example](https://github.com/Unity-Technologies/datasetinsights/blob/master/tests/evaluation_metrics/test_average_relative_error.py).

pyproject.toml
```diff
@@ -44,7 +44,10 @@ pandas = "^1.0.1"
 tqdm = "^4.45.0"
 dask = {extras = ["complete"], version = "^2.14.0"}
 dash = "1.12.0"
-tensorflow = "^2.2.0"
+tensorflow = "^1.13.1"
```
BlairLee: Why do we need to downgrade the version?

BlairLee:
Please solve linting issues and pytest errors.

pyproject.toml
JonathanLeban: So I have a problem regarding this. If I upgrade the version, I can't load my model the way I want. But if the version is not at least 2, the groceries.real.py file prints out errors....

adason: We have decided to have this repo using tensorflow 2.X. Can you make sure to update your tensorflow model so that it's compatible with this requirements?

PR #42, opened 2020-08-14
import fix

sanjayuconn:
# Peer Review Information

- Removed the synthetic import in init and fixed corresponding imports

# Pull Request Check List

### DIFF
--- a/datasetinsights/datasets/__init__.py
+++ b/datasetinsights/datasets/__init__.py
@@ -3,7 +3,6 @@ from .cityscapes import Cityscapes
 from .coco import CocoDetection
 from .groceries_real import GoogleGroceriesReal, GroceriesReal
 from .nyudepth import NyuDepth
-from .synthetic import SynDetection2D, read_bounding_box_2d
 
 __all__ = [
     "Cityscapes",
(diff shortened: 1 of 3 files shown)


notebooks/SynthDet_bounding_box_visualization.ipynb
```diff
@@ -21,7 +21,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 2,
```
BlairLee, line 24: Please clear up all the outputs in the notebook.

datasetinsights/datasets/__init__.py
```diff
@@ -12,6 +11,4 @@
     "GroceriesReal",
     "GoogleGroceriesReal",
     "NyuDepth",
-    "SynDetection2D",
```
BlairLee, line 15: [mustfix] If we remove `SynDetection2D`, it cannot find `SynDetection2D` class in `Dataset.find()`.

sanjayuconn: Yes I agree. We need to move usim.py under the datasets in this case, usim.py and synthetic.py have cyclic dependency.

adason:
@sanjayuconn  We should remove `from datasetinsights.io.usim import Downloader, download_manifest` in `Synthetic.py` and remove calls that refer to download data for unity simulation. This code path is already dead after @Saurav-D makes `SynthDetection2D` to directly download from public http url.

sanjayuconn:
Will send new PR.

PR #46, opened 2020-08-18
Kubeflow Pipelines for SynthDet Release

adason:
# Peer Review Information

This PR simplified multiple pipeline with small input parameter variations. It uses a few reusable operations and compose different pipelines. Currently, the pipeline will not be working since we most refactoring of cli is not completed. This review simply focus on the interface design. We should have separate PRs to bug fixes these pipelines one the new cli is implemented.

The compiled yaml file are included in git history to allow users to have directly access to these pipelines from github.

evaluate_the_model
train_on_real_world_dataset
train_on_synthdet_sample
train_on_synthetic_and_real_dataset
train_on_synthetic_dataset_unity_simulation

# Pull Request Check List


kubeflow/pipelines.py
```diff
@@ -0,0 +1,445 @@
...
+        val_data=val_data,
+        tb_log_dir=tb_log_dir,
+        checkpoint_dir=checkpoint_dir,
+        volume=download.pvolumes[DATA_PATH],
+        memory_limit=memory_limit,
+        num_gpu=num_gpu,
+        gpu_type=gpu_type,
+    )
+
+    return train
```
AdamPalmarUnity: What is the purpose of the return calls here and in the other dsl.pipelines?

kubeflow/pipelines.py
```diff
@@ -0,0 +1,445 @@
...
+    evaluate.apply(gcp.use_gcp_secret("user-gcp-sa"))
+
+    return evaluate
+
+
+@dsl.pipeline(
+    name="Train on the SynthDet sample",
+    description="Train on the SynthDet sample",
+)
+def train_on_synthdet_sample(
```
AdamPalmarUnity, line 203: Is there a difference between this dsl.pipeline and the other pipelines besides the default arguments?

kubeflow/pipelines.py
```diff
@@ -0,0 +1,445 @@
...
+    )
+
+    return train
+
+
+@dsl.pipeline(
+    name="Train on Synthetic + Real World Dataset",
+    description="Train on Synthetic + Real World Dataset",
+)
+def train_on_synthetic_and_real_dataset(
```
AdamPalmarUnity, line 343: Should this pipeline include downloading of the synthetic dataset? Currently it only downloads the groceries real dataset from what it looks like.

kubeflow/pipelines.py
```diff
@@ -0,0 +1,445 @@
...
+            "datasetinsights",
+            "train",
+        ]
+    else:
+        command = ["datasetinsights", "train"]
+
+    arguments = [
+        f"--config={config}",
+        f"--train-data={train_data}",
+        f"--val-date={val_data}",
```
AdamPalmarUnity: Typo

kubeflow/pipelines.py
```diff
@@ -0,0 +1,445 @@
...
+    """ Create a Kubeflow ContainerOp to train an estimator.
+
+    Args:
+        docker (str): Docker image registry URI.
+        config (str): Path to estimator config file.
+        train_data (str): Path to train dataset directory.
+        val_data (str): Path to val dataset directory.
+        checkpoint_file (str): Path to an estimator checkpoint file.
+            If specified, model will resume from previous checkpoints.
+        tb_log_dir (str): Path to tensorbload log directory.
```
AdamPalmarUnity: Typo

kubeflow/pipelines.py
adason: It's not used in these pipelines. It simply prevent `flake8` error as train variable is defined but not used. I will remove these return.

kubeflow/pipelines.py
adason: This pipeline actually does not require Synthetic Dataset. It should consume a model train on a different pipeline (which uses synthtic dataset)

kubeflow/pipelines.py
adason: Good catch. Fixed now.

kubeflow/pipelines.py
adason: fixed

kubeflow/pipelines.py
adason: They are not very different. We can probably make a few pipeline and share some of the components, but this might be over-engineering these pipelines. Using default arguments makes it easier to demonstrate different training/evaluation procedures.

kubeflow/pipelines.py
AdamPalmarUnity: The function name is a bit misleading then. But its a minor thing. If the model is doing fine tuning then it should say so i think.

AdamPalmarUnity (approved):

PR #47, opened 2020-08-18
HTTPDownloader

Saurav-D:
# Peer Review Information

I've added:
- `HTTPDownloader` class which extends `DatasetDownloader` class
- `checksum_file` option in the download cli.

# Pull Request Check List


datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    def get_checksum_from_file(filepath):
+
+        if filepath.startswith(("http://", "https://")):
+            with tempfile.TemporaryDirectory() as tmp:
+                checksum_file_path = os.path.join(tmp, "checksum.txt")
+                download_file(source_uri=filepath, dest_path=checksum_file_path)
+                return HTTPDownloader.read_checksum_from_txt(checksum_file_path)
+
+        elif filepath.startswith("/"):
```
AdamPalmarUnity: You might want to do `os.path.isfile(fname) ` instead. It's possible that the user specifies a relative path to the file based on where the user is executing the command.

Saurav-D: Thanks, will do.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,66 @@
...
+import tempfile
+import zipfile
+
+from datasetinsights.io.download import download_file, validate_checksum
+from datasetinsights.io.downloader.base import DatasetDownloader
+from datasetinsights.io.exceptions import ChecksumError
+
+logger = logging.getLogger(__name__)
+
+DATASETS = ["synthetic", "groceries_real"]
```
AdamPalmarUnity: What will these variables be used for?

Saurav-D: Forgot to remove them. Will remove.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,64 @@
...
+    def get_checksum_from_file(filepath):
+
+        if filepath.startswith(("http://", "https://")):
+            with tempfile.TemporaryDirectory() as tmp:
+                checksum_file_path = os.path.join(tmp, "checksum.txt")
+                download_file(source_uri=filepath, dest_path=checksum_file_path)
+                return HTTPDownloader.read_checksum_from_txt(checksum_file_path)
+
+        elif os.path.isfile(filepath):
+            checksum_file_path = filepath
```
AdamPalmarUnity: [minor] pass the filepath directly to the staticmethod

datasetinsights/datasets/groceries_real.py
```diff
@@ -126,19 +110,13 @@ def __init__(
...
-                f"Available versions are: {valid_versions}"
-            )
         logger.info(
             f"Using split {split} and version {version} for groceries real "
             "dataset."
         )
 
         self.version = version
-        self.root = os.path.join(data_root, self.LOCAL_PATH)
+        self.root = data_root
```
AdamPalmarUnity: I think it makes sense to rename this to data_path or data_directory or something along those lines. As well as the data_root constructor argument

tests/test_http_downloader.py
```diff
@@ -0,0 +1,61 @@
...
+import pytest
+
+from datasetinsights.io.downloader.http_downloader import HTTPDownloader
+from datasetinsights.io.exceptions import ChecksumError
+
+
+def test_read_checksum_from_txt():
+    with tempfile.NamedTemporaryFile(mode="w+") as tmp:
+        tmp.write("123456")
+        tmp.read()
```
AdamPalmarUnity: I think it makes more sense to do a tmp.flush() to force the write to the file instead of tmp.read()

tests/test_http_downloader.py
```diff
@@ -0,0 +1,61 @@
...
+        assert HTTPDownloader.get_checksum_from_file(tmp.name) == 123456
+
+
+@patch("datasetinsights.io.downloader.http_downloader.download_file")
+@patch.object(HTTPDownloader, "read_checksum_from_txt")
+def test_get_checksum_from_file(
```
AdamPalmarUnity: Unit test should follow the pattern. 
arrange 
act 
assert 

This will make the tests easier to understand. This unit test does multiple things and should instead be refactored into multiple tests.

tests/test_http_downloader.py
```diff
@@ -0,0 +1,61 @@
...
+    mock_read_checksum_from_txt, mock_download_file
+):
+    filepath_1 = "http://some/path"
+    filepath_2 = "https://some/path"
+    filepath_3 = "some/wrong/path"
+
+    mock_read_checksum_from_txt.return_value = 123456
```
AdamPalmarUnity, line 26: This defeats the purpose of the test. Mocking the function you are testing does not validate the behavior it just tests if mocking a function works.

tests/test_http_downloader.py
```diff
@@ -0,0 +1,61 @@
...
+    HTTPDownloader.get_checksum_from_file(filepath_1)
+    mock_download_file.assert_called_once()
+    mock_read_checksum_from_txt.assert_called_once()
+
+    assert HTTPDownloader.get_checksum_from_file(filepath_2) == 123456
+
+    with pytest.raises(ValueError):
+        HTTPDownloader.get_checksum_from_file(filepath_3)
+
+    with tempfile.NamedTemporaryFile() as tmp:
```
AdamPalmarUnity: This case is already being tested in test_read_checksum_file_from_txt

datasetinsights/io/downloader/base.py
```diff
@@ -19,15 +19,16 @@ def _find_downloader(source_uri):
     """
     protocols = "|".join(_registry.keys())
     pattern = re.compile(f"({protocols})")
+
     protocol = pattern.findall(source_uri)
-    if protocol:
+
+    if source_uri.startswith(("https://", "http://")):
```
adason, line 24: If protocol can't be matched, this will fail. `None string have no method called start_with...`. This should be fixed and add unittest tests for such cases.

datasetinsights/io/downloader/base.py
```diff
@@ -48,6 +49,9 @@ def create_downloader(source_uri, **kwargs):
 
 
 class DatasetDownloader(ABC):
+    def __init__(self, **kwargs):
```
adason, line 51: Should be removed. This is not required.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,105 @@
...
+logger = logging.getLogger(__name__)
+
+
+class HTTPDownloader(DatasetDownloader, protocol="http://"):
+    """ This class is used to download data from any HTTP or HTTPS public url
+        and perform function such as downloading the dataset and checksum
+        validation.
+    """
+
+    def __init__(self, **kwargs):
```
adason: Not required to have __init__ method.

datasetinsights/commands/download.py
```diff
@@ -79,8 +79,15 @@ def convert(self, value, param, ctx):
     help="Unity Simulation access token. "
     "This will override synthetic datasets source-uri for Unity Simulation",
 )
+@click.option(
+    "--checksum-file",
+    type=str,
+    default=None,
+    help="Dataset checksum text file. "
```
adason: Since `checksum-file` can either be a http url or file in the local filesystem, you should make this clear in the help message.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,105 @@
...
+                validate_checksum(dataset_path, checksum)
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Deleting the downloaded file.")
+                os.remove(dataset_path)
+                raise e
+
+        HTTPDownloader.unzip_file(dataset_path, output)
+
+    @staticmethod
+    def get_checksum_from_file(filepath):
```
adason: This method seems to be generic enough that can be used for other `DatasetDownloader`. Maybe it should be moved to the same module where we compute checksums?

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,105 @@
...
+                return HTTPDownloader.read_checksum_from_txt(checksum_file_path)
+
+        elif os.path.isfile(filepath):
+            return HTTPDownloader.read_checksum_from_txt(filepath)
+
+        else:
+            raise ValueError(f"Can not get checksum from path: {filepath}")
+
+    @staticmethod
+    def read_checksum_from_txt(filepath):
```
adason: This method seems to be generic enough that can be used for other `DatasetDownloader`. Maybe it should be moved to the same module where we compute checksums?

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,105 @@
...
+            checksum = HTTPDownloader.get_checksum_from_file(checksum_file)
+            try:
+                logger.debug("Validating checksum!!")
+                validate_checksum(dataset_path, checksum)
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Deleting the downloaded file.")
+                os.remove(dataset_path)
+                raise e
+
+        HTTPDownloader.unzip_file(dataset_path, output)
```
adason: [nit] use `self.unzip_file(...)` to save some typing. This will also make it easier if we have to change the class name `HTTPDownloader` in the future.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,105 @@
...
+                                 can be HTTP or HTTPS url or local path.
+
+        Raises:
+            ChecksumError: This will raise this error if checksum doesn't
+                           matches
+
+        """
+
+        dataset_path = os.path.join(output, "dataset.zip")
+        download_file(source_uri, dataset_path)
```
adason, line 42: What happened if the file download was not successful? Should we catch error and re-raise exceptions?

tests/test_create_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+from datasetinsights.io.downloader.unity_simulation import (
+    UnitySimulationDownloader,
+)
+
+
+@pytest.mark.parametrize(
+    "source_uri", ["http://", "https://"],
+)
+def test_create_downloader_http_downloader(source_uri):
+    # arrange
```
adason: ??? This comment seems out of place.

tests/test_create_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+
+    # act
+    downloader = create_downloader(source_uri=source_uri)
+
+    # assert
+    assert isinstance(downloader, HTTPDownloader)
+
+
+def test_create_downloader_unity_simulation_downloader():
```
adason, line 22: Why test create_downloader when `_find_downloader` is tested against different cases? What about cases when _find_downloader can find a downloader for a given source_uri? What about cases when source_uri is none?

tests/test_http_downloader.py
```diff
@@ -0,0 +1,145 @@
...
+    )
+
+    # assert
+    mock_download_file.assert_called_once()
+    mock_unzip.assert_called_once_with(dataset_path, output)
+    mock_get_checksum_from_file.assert_called_once()
+    mock_validate_check_sum.assert_called_once()
+
+
+@pytest.mark.parametrize(
```
adason: Why do we need different parameter when we have `mock_validate_checksum.side_effect = ChecksumError`. In this does it matter to vlaidate multiple combination of `source_uri` and `checksum_file` here, if `validate_checksum` method is always going to raise?

tests/test_http_downloader.py
```diff
@@ -0,0 +1,145 @@
...
+        HTTPDownloader.get_checksum_from_file(filepath)
+
+    # assert
+    mock_download_file.assert_not_called()
+    mock_read_checksum_from_txt.assert_not_called()
+
+
+@pytest.mark.parametrize(
+    "source_uri",
+    ["http://some/path", "https://some/path", "some/checksum_file.txt"],
```
adason: Why do you need a `some/checksum_file.txt` for `source_uri`? Also, the name of this method seems to suggest it tries to run logic without checksum. Why do we need to have both `source_uri=http, https` when the logic of handling either http/https source uri is already covered by different use cases? Ideally, user should be able to provide any random string for `source_uri` and be able to test the behavior when `checksum_file` is not provided.

tests/test_http_downloader.py
```diff
@@ -0,0 +1,145 @@
...
+    with tempfile.NamedTemporaryFile(mode="w+") as tmp:
+        tmp.write("123456")
+        tmp.flush()
+        # act
+        checksum = HTTPDownloader.get_checksum_from_file(tmp.name)
+        # assert
+        assert checksum == 123456
+
+
+@pytest.mark.parametrize("filepath", ["http://some/path", "https://some/path"])
```
adason: It will be nice to add  a few test cases when a given `filepath` will raise `ValueError`.

datasetinsights/io/downloader/http_downloader.py
Saurav-D: That's been handled in `download_file` function. It'll raise `DownloadError` if unsuccessful.

datasetinsights/io/downloader/http_downloader.py
AdamPalmarUnity: This will fail if the constructor is called with an argument like access_token.

tests/test_create_downloader.py
AdamPalmarUnity: The arrange part could be removed

tests/test_create_downloader.py
AdamPalmarUnity: Good point, there is no test for create_downloader with input None. I don't understand what you mean by the two first questions though.

tests/test_http_downloader.py
AdamPalmarUnity: The test below does this.

tests/test_http_downloader.py
AdamPalmarUnity: The some/checksum_file.txt can be removed from the test case. The test is about not providing a checksum file and validating that the download and unzip is called properly

tests/test_http_downloader.py
AdamPalmarUnity: The importance of the test is the following

```
    mock_get_checksum_from_file.assert_called_once()
    mock_download_file.asert_called_once()
    mock_remove.assert_called_once()
    mock_unzip_file.assert_not_called()
```

tests/test_http_downloader.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # act
+    with pytest.raises(ChecksumError):
+        downloader.download(
+            source_uri=source_uri, output=output, checksum_file=checksum_file
+        )
+
+    # assert
+    mock_get_checksum_from_file.assert_called_once()
+    mock_download_file.asert_called_once()
```
AdamPalmarUnity: Typo here.

Saurav-D: What's the typo?

datasetinsights/io/downloader/base.py
AdamPalmarUnity: Are you sure. its the source_uri that is being called not protocol

tests/test_http_downloader.py
AdamPalmarUnity: I don't understand your comment on what is wrong with the test besides some/checksum_file.txt

datasetinsights/io/downloader/http_downloader.py
Saurav-D: Moved to `download.py` where checksum is computed.

datasetinsights/io/downloader/http_downloader.py
Saurav-D: Moved to `download.py` where checksum is computed.

tests/test_http_downloader.py
Saurav-D: I think @adason mean that we don't need to test for both `http` and `https` since this test doesn't check whether it handles `http` and `https` but it's about whether download and unzip is called properly.

tests/test_http_downloader.py
```diff
@@ -0,0 +1,101 @@
...
+
+    # act
+    downloader.download(source_uri=source_uri, output=output)
+
+    # assert
+    mock_download_file.assert_called_once()
+    mock_unzip.assert_called_once_with(dataset_path, output)
+
+
+@pytest.mark.parametrize(
```
Saurav-D: Here also we don't need to parameterize `source_uri`

tests/test_http_downloader.py
AdamPalmarUnity: asert_called_once

tests/test_http_downloader.py
AdamPalmarUnity: Ok in that case we can remove the parametrize method and just have one source_uri

datasetinsights/io/downloader/base.py
AdamPalmarUnity: I'll remove it from here

tests/test_create_downloader.py
AdamPalmarUnity: I don't think it makes sense to unit test _find_downloader as it is a private function

tests/test_http_downloader.py
AdamPalmarUnity: I'll remove the parametrize for this test

tests/test_http_downloader.py
AdamPalmarUnity: Are there any cases you have in mind?

datasetinsights/io/downloader/base.py
adason: You are correct. 👍

datasetinsights/io/downloader/http_downloader.py
adason: `__init__` method should lookup it's parent class `DatasetDownloader` which should be able to catch access_token in  **kwargs?

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+                validate_checksum(dataset_path, checksum)
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Deleting the downloaded file.")
+                os.remove(dataset_path)
+                raise e
+
+        self.unzip_file(dataset_path, output)
+
+    @staticmethod
+    def unzip_file(filepath, destination):
```
adason: @AdamPalmarUnity @Saurav-D 

I just realized that if we push the responsibility of unzipping files to `DatasetDownloader`, we actually make a very strong assumption about the file that should be downloaded. In fact, don't know what type of compressed file we have for a given `source_uri`. It could download an tfdata `protobuf` file, `hdf5` file, `tar.gz` file, ... 

**We should not make unzip part of the responsibility of HTTPDownloader**

tests/test_create_downloader.py
adason: We should have unittest for ANY function that preform non-trivial operations, whether it's public or private function. The logic inside `create_downloader` seems trivial however `_find_downloader` is non trivial.  

Typo in my previous comment. Correction: 

> What about cases when `_find_downloader` can't find a `DatasetDownloader` from a given source_uri? What about cases when source_uri is none?

datasetinsights/io/downloader/base.py
adason: Actually, from the comments in `HTTPDownloader`, we should keep this to catch all arguments that was passed to constructor.  By default, when users creates a new Downloader without `__init__` method, it's parent class `DatasetDownloader` should handle such cases.

tests/test_http_downloader.py
adason: A few invalid cases: ["zvsssdfsdg", ":path with space", ""]

Saurav-D: Added [here](https://github.com/Unity-Technologies/datasetinsights/pull/47/commits/93e30c6571d0fb0db175dacc03e7cfb072fc5a0e)

datasetinsights/io/downloader/http_downloader.py
Saurav-D: Then whose responsibility should be to unzip the dataset file. Since Dataset class expects just a data_directory which contains all the files (annotations, images etc.) when should unzip happen and where? Also, do we want to support different dataset file types?

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+            checksum = get_checksum_from_file(checksum_file)
+            try:
+                logger.debug("Validating checksum!!")
+                validate_checksum(dataset_path, checksum)
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Deleting the downloaded file.")
+                os.remove(dataset_path)
+                raise e
+
+        self.unzip_file(dataset_path, output)
```
BlairLee: [must fix] I've merged this branch to my local branch. I figured out that this would unzip it to a subfolder. 
For example, if we set output = "/data/groceries/v3". It would download a "dataset.zip" in the "/data/groceries/v3" and unzip it to "/data/groceries/v3/v3".

Saurav-D: Yes, that's because groceries dataset zip file contains a directory `v3` which in turn contains the dataset files. I don't think it should be the responsibility of downloader to find the files or get rid of directories, instead all the dataset files should be zipped directly instead of a zipping a directory that contains those files.

BlairLee: I will zip the GroceriesReal directly and update on GCS.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,71 @@
...
+        """ This method is used to download the dataset from HTTP or HTTPS url.
+
+        Args:
+            source_uri (str): This is the downloader-uri that indicates where
+                              the dataset should be downloaded from.
+
+            output (str): This is the path to the directory where the download
+                          will store the dataset.
+
+            checksum_file (str): This is path of the txt file that contains
```
BlairLee, line 31: Are we going to provide checksum files of the datasets in the public GCS bucket?

datasetinsights/io/downloader/base.py
AdamPalmarUnity: I'll keep it here then and remove it from the other http_downloader

tests/test_create_downloader.py
AdamPalmarUnity: Added these two tests that test the interface.


```
def test_create_downloader_invalid_input():
    # arrange
    source_uri = "invalid_protocol://"
    # assert
    with pytest.raises(ValueError):
        # act
        create_downloader(source_uri=source_uri)


def test_create_downloader_none_input():
    # arrange
    source_uri = None
    # assert
    with pytest.raises(TypeError):
        # act
        create_downloader(source_uri=source_uri)
```

AdamPalmarUnity: I think we should focus on testing the interface instead of the implementation

datasetinsights/io/downloader/http_downloader.py
AdamPalmarUnity: In that case i think it makes sense to refactor the unzipping/ decompressing into another class that is injected into this class. Then we can also reuse that in the GCSDownloader. We could call it decompressor. It could also be a function.

tests/test_create_downloader.py
AdamPalmarUnity: What kind of cases is missing?

datasetinsights/io/downloader/http_downloader.py
AdamPalmarUnity: To answer Saurav's question i still think the unzipping/decompressing can happen in the HTTPDownloader but the code for how to do it should not live here.

datasetinsights/io/downloader/http_downloader.py
Saurav-D: Yes.

datasetinsights/io/compression.py
```diff
@@ -0,0 +1,97 @@
...
+
+        """
+        destination = os.path.join(
+            destination, os.path.splitext(os.path.basename(filepath))[0]
+        )
+        with open(destination, "wb") as out_f, gzip.GzipFile(filepath) as zip_f:
+            out_f.write(zip_f.read())
+
+
+def get_file_type_from_filepath(filepath):
```
AdamPalmarUnity: Would probably make this private.

AdamPalmarUnity (dismissed):

tests/test_compression.py
```diff
@@ -0,0 +1,45 @@
...
+
+from datasetinsights.io.compression import (
+    GZipCompression,
+    TarFileCompression,
+    ZipFileCompression,
+    compression_factory,
+)
+
+
+@patch("datasetinsights.io.compression._get_file_extension_from_filepath")
```
AdamPalmarUnity: Why are you mocking out the private function?

Saurav-D: Because there is no actual file to get the extension. Is there any other way?

Saurav-D: Also maybe there is no need to make that method private.

datasetinsights/io/compression.py
```diff
@@ -0,0 +1,99 @@
...
+import os
+import tarfile
+import zipfile
+
+import filetype
+
+logger = logging.getLogger(__name__)
+
+
+class ZipFileCompression:
```
AdamPalmarUnity: After looking around abit i found this. 

import shutil
This function could replace the different compressors.
shutil.unpack_archive() 

It takes a filename , output dir and format. If you use the _get_file_extension_from_filepath(filepath) function plus this we can saw our self from reimplementing the different formats

tests/test_compression.py
AdamPalmarUnity: I'll push some tests and we can take a look if they make sense

datasetinsights/io/compression.py
Saurav-D: Okay, I'll implement this.

AdamPalmarUnity: Remember to pull i just pushed some tests maybe they are not needed at this point, if this is implemented. Feel free to remove them if need be.

Saurav-D: After reading the documentation, which states: `format is the archive format: one of “zip”, “tar”, “gztar”, “bztar”, or “xztar”`. That is they only support these 5 formats rest of the formats we'll have to register. Also, out of these 5 only 2 are most popular which means apart from these 2 we'll have to anyway implement other formats. Libraries like pytorch as well have their own [implementation](https://github.com/pytorch/vision/blob/dcfcc8675393749e852d8956b580b1660dfbea2d/torchvision/datasets/utils.py#L214).

Saurav-D: Will implement this in new PR.

datasetinsights/io/download.py
```diff
@@ -122,3 +124,40 @@ def _crc32_checksum(filepath):
         checksum = zlib.crc32(f.read())
 
     return checksum
+
+
+def get_checksum_from_file(filepath):
+    """ This method return checksum of the file whose filepath is given.
+
+    Args:
+        filepath (str): Path of the checksum file.
```
adason, line 133: Should add note to explain filepath can be either `http url` or `local_path`

datasetinsights/io/download.py
```diff
@@ -122,3 +124,40 @@ def _crc32_checksum(filepath):
...
+def read_checksum_from_txt(filepath):
+    """ This method reads checksum from a txt file and returns it.
+
+    Args:
+        filepath (str): Local filepath of the checksum file.
+
+    """
+    with open(filepath) as file:
+        checksum = file.read()
+    return int(checksum)
```
adason: [mustfix] checksum isn't always integer.

datasetinsights/io/download.py
```diff
@@ -122,3 +124,40 @@ def _crc32_checksum(filepath):
...
+    else:
+        raise ValueError(f"Can not get checksum from path: {filepath}")
+
+
+def read_checksum_from_txt(filepath):
+    """ This method reads checksum from a txt file and returns it.
+
+    Args:
+        filepath (str): Local filepath of the checksum file.
+
```
adason, line 162: Missing docstring for the return value of this method? Also, is this method going to be used as public method? Maybe we can make it private?

datasetinsights/io/downloader/base.py
```diff
@@ -12,22 +12,22 @@ def _find_downloader(source_uri):
         Args:
             source_uri: URI of where this data should be downloaded.
 
-            Returns: The dataset downloader class that is
-             registered with the source-uri protocol
-
+        Returns: The dataset downloader class that is registered with the
```
adason, line 15: Return docstring should follow this pattern

```
Returns:
    type: description
```

see this [doc](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for more examples

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -0,0 +1,53 @@
...
+    get_checksum_from_file,
+    validate_checksum,
+)
+from datasetinsights.io.downloader.base import DatasetDownloader
+from datasetinsights.io.exceptions import ChecksumError
+
+logger = logging.getLogger(__name__)
+
+
+class HTTPDownloader(DatasetDownloader, protocol="http://"):
```
adason: [mustfix] We agree to change this to `HTTPDatasetDownloader`

pyproject.toml
```diff
@@ -46,6 +46,7 @@ dask = {extras = ["complete"], version = "^2.14.0"}
 dash = "1.12.0"
 tensorflow = "^2.2.0"
 click = "^7.1.2"
+filetype = "^1.0.7"
```
adason: Why is this package required in this review? Is it used as part of this review?

tests/test_create_downloader.py
adason: The above two cases are good to add. You will want to test both interface and implementation in unittests. We will not be able to cover all edge cases in one pass, but as we work on the repo and fixing bugs, we should continue adding new cases that will cause a bug.

adason (changes requested):

pyproject.toml
Saurav-D: Removed.

datasetinsights/io/downloader/http_downloader.py
Saurav-D: will do.

datasetinsights/io/download.py
Saurav-D: We can make it private, I have one question though, should we test private methods?

adason (approved):

PR #50, opened 2020-08-19
GCSDownloader

sanjayuconn:
# Peer Review Information
- design doc.(https://docs.google.com/document/d/13QtPZ3_Hlb5de36_ZJE-VlLm9-Ycg2_Syp0vEkPIkJ0/edit#)

What to review: 
- Download file and directory
- Checksum validation
- Fixed existing dependency on GCSClient and updated docstring
- Upload and its dependency 
- Unit testing

Note: 
- Please review checkpoint.py carefully. I have refactor the GCSEstimatorWriter's  download and upload code and removed bucket and key split call. 

# Pull Request Check List


datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+
+class GCSDownloader(DatasetDownloader, protocol="gs://"):
+    """ This class is used to download data from GCS
+    """
+
+    def __init__(self, **kwargs):
+        """
+
+        Args:
+            access_token: Access token to be used to authenticate to
```
AdamPalmarUnity: I don't think this arg is required.

datasetinsights/io/gcs.py
```diff
@@ -101,3 +109,52 @@ def download_file_from_gcs(cloud_path, local_path, filename, use_cache=True):
 
     # TODO(YC) Should run file checksum before return.
     return local_filepath
+
+
+def download_folder_from_gcs(cloud_path, local_path):
```
AdamPalmarUnity: I would expect the below code to be in GCSDownloader. Together with download_file_from_gcs to be in the GCSDownloader or the GCSClient. The reason being the if the GCSClient would evolve to take a service account in the constructor this is now not possible as the functions in this module depends on the GCSClient but assumes it takes no arguments.

datasetinsights/io/gcs.py
```diff
@@ -101,3 +109,52 @@ def download_file_from_gcs(cloud_path, local_path, filename, use_cache=True):
...
+
+    Examples:
+        >>> cloud_path = "gs://bucket/folder"
+        >>> local_path = "/tmp/folder"
+        >>> filename = "file.txt"
+        >>> download_file_from_gcs(cloud_path, local_path, filename)
+        # download file gs://bucket/folder/file.txt to /tmp/folder/file.txt
+    """
+    bucket, prefix = gcs_bucket_and_path(cloud_path)
+    client = GCSClient().client
```
AdamPalmarUnity: Why not just instantiate the Client instead if that's what is needed? 
This also goes against the principal of depedency injection. If a function or a method is dependent on functionality this should be injected. This will also make it easier to test.
If we at some point decide to implement a new GCSClient that can donwload things in parallel or invoke gsutils instead, its good to have that encapsulated in the class and not spread over multiple functions.

datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+        Args:
+            source_uri: This is the downloader-uri that indicates where on
+                GCS the dataset should be downloaded from.
+                The expected source-uri follows these patterns
+                usim: gs://bucket/folder
+
+            output: This is the path to the directory
+                where the download will store the dataset.
+
+        """
```
AdamPalmarUnity, line 25: There is also the case if the source_uri points to a zip file that contains the dataset.

datasetinsights/io/gcs.py
```diff
@@ -1,11 +1,19 @@
...
+from datasetinsights.io.download import validate_checksum
+from datasetinsights.io.exceptions import ChecksumError, DownloadError
+
 logger = logging.getLogger(__name__)
 
+MD5 = "MD5"
+REPAD = "=="
+
 
 class GCSClient:
```
AdamPalmarUnity, line 17: @adason Why are the functions below not a part of the GCSClient?
gcs_bucket_and_path, copy_folder_to_gcs, download_file_from_gcs

datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+        download_folder_from_gcs(source_uri, output)
```
AdamPalmarUnity: Could also be done with gsutil as an alternative that we should consider. Then we dont need to worry about checksums and optimizing download speed and writing a lot of tests

```
import subprocess

subprocess.check_call(['gsutil', '-m', 'cp','-r', source_uri, output_directory])
```

datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+            unity simulation for downloading the dataset
+        """
+
+    def download(self, source_uri, output, **kwargs):
+        """
+
+        Args:
+            source_uri: This is the downloader-uri that indicates where on
+                GCS the dataset should be downloaded from.
+                The expected source-uri follows these patterns
```
AdamPalmarUnity, line 20: It could also be a single file that is zipped

datasetinsights/io/downloader/gcs_downloader.py
sanjayuconn: Its good but have one problem. It creates an additional directory: Ex:
gsutil cp -r gs://my-bucket/data dir
- download from gs://my-bucket/data to be  dir/data/a/b/c. 
- Local directory start from data, which we do not want. It should be something like that dir/a/b/c
- We need to move files one layer up

datasetinsights/io/download.py
```diff
@@ -122,3 +126,13 @@ def _crc32_checksum(filepath):
...
+
+
+def _md5_checksum(filename):
+    """ Calculate the checksum of a file using MD5.
+    """
+    md5 = hashlib.md5()
+    with open(filename, "rb") as f:
+        for chunk in iter(lambda: f.read(4096), b""):
+            md5.update(chunk)
+    return base64.b64encode(md5.digest()).decode("utf-8")
```
adason: Why is this checksum base64 encoded and decoded to "utf-8"? Why we can't use the md5 checksum as is?

datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,30 @@
...
+    """ This class is used to download data from GCS
+    """
+
+    def __init__(self, **kwargs):
+        """
+
+        Args:
+            access_token: Access token to be used to authenticate to
+            unity simulation for downloading the dataset
+        """
```
adason: `__init__` method should initialize a GCS client that have the right credential (read from default credential location or environment variable). In the future, we might want to support initialization using a path to GCS service account as well as optional parameter such as GCP project.

https://googleapis.dev/python/storage/latest/client.html#module-google.cloud.storage.client

datasetinsights/io/downloader/gcs_downloader.py
adason: We should use GCP python [client](https://googleapis.dev/python/storage/latest/client.html#module-google.cloud.storage.client) here. Using subprocess call is usually a hack when python library is not available.

datasetinsights/io/gcs.py
```diff
@@ -1,11 +1,19 @@
...
 from pathlib import Path
 
 from google.cloud.storage import Client
 
+from datasetinsights.io.download import validate_checksum
+from datasetinsights.io.exceptions import ChecksumError, DownloadError
+
 logger = logging.getLogger(__name__)
 
+MD5 = "MD5"
```
adason: Why have a constant that is exactly the same as string?

datasetinsights/io/gcs.py
```diff
@@ -101,3 +109,52 @@ def download_file_from_gcs(cloud_path, local_path, filename, use_cache=True):
...
+            blob.download_to_filename(dst_file_name)
+        except DownloadError as e:
+            logger.info(
+                f"The request download from {prefix} -> {dst_file_name} can't "
+                f"be completed."
+            )
+            raise e
+        expected_checksum = blob.md5_hash
+        if expected_checksum:
+            expected_checksum += REPAD
```
adason: Why do we need to append "==". This and base64 encode/decode seems to be something less intuitive. We should understand how to get the raw value MD5 cheksums and validate this value directly.

datasetinsights/io/gcs.py
```diff
@@ -101,3 +109,52 @@ def download_file_from_gcs(cloud_path, local_path, filename, use_cache=True):
...
+        >>> filename = "file.txt"
+        >>> download_file_from_gcs(cloud_path, local_path, filename)
+        # download file gs://bucket/folder/file.txt to /tmp/folder/file.txt
+    """
+    bucket, prefix = gcs_bucket_and_path(cloud_path)
+    client = GCSClient().client
+    bucket = client.get_bucket(bucket)
+    blobs = bucket.list_blobs(prefix=prefix)
+    for blob in blobs:
+        blob_name = blob.name
```
adason: Why not reuse `download_file_from_gcs` or refactor as necessary?

datasetinsights/io/gcs.py
```diff
@@ -101,3 +109,52 @@ def download_file_from_gcs(cloud_path, local_path, filename, use_cache=True):
 
     # TODO(YC) Should run file checksum before return.
     return local_filepath
+
+
+def download_folder_from_gcs(cloud_path, local_path):
+    """Helper method to download list of files from GCS
```
adason: This method tries to do multiple things. Might be helpful to break it down to smaller unit of functions ant make it easier for tests.

datasetinsights/io/gcs.py
adason: +1

adason (changes requested):

datasetinsights/io/downloader/gcs_downloader.py
sanjayuconn: Do we have any identifier from cli that indicates whether source GCS location is file or directory? If no then we include flag = file/dir and accordingly it will call two seperate flow to download from blob and blobs.

AdamPalmarUnity: I think it should be inferred from the source_uri instead of passing it trough the cli

datasetinsights/io/downloader/gcs_downloader.py
AdamPalmarUnity: From the comment i don't see why its a hack can you elaborate on why this would be the case?

AdamPalmarUnity: If the extra dir is inconvenient, you could use gsutil 'rsync' instead i think

datasetinsights/io/downloader/gcs_downloader.py
sanjayuconn: There are so many file compression, what if there is no indicator for file type(.zip or .*) then how to distinguish. Ex: gs://my-bucket/data, suppose data is file here, how to infer?
- Also directory name could be(. separated) ex: gs://my-bucket/data.tmp

datasetinsights/io/gcs.py
```diff
@@ -1,25 +1,34 @@
...
 
 class GCSClient:
     def __init__(self, **kwargs):
         """ Initialize a client to google cloud storage (GCS).
         """
         self.client = Client(**kwargs)
 
-    def download(self, bucket_name, object_key, localfile):
+    def download(
+        self, local_path, bucket_name=None, key=None, url=None, filename=None
```
AdamPalmarUnity: What is this filename parameter for?

datasetinsights/io/downloader/gcs_downloader.py
```diff
@@ -0,0 +1,27 @@
+from datasetinsights.io.downloader.base import DatasetDownloader
+from datasetinsights.io.gcs import GCSClient
+
+
+class GCSDownloader(DatasetDownloader, protocol="gs://"):
```
AdamPalmarUnity: Should be renamed to GCSDatasetDownloader

datasetinsights/io/downloader/gcs_downloader.py
AdamPalmarUnity: You have to download the file first to know if its compressed or not

AdamPalmarUnity: So you can tell if its a directory or a file i suppose

AdamPalmarUnity: I think @Saurav-D is working on a decompressor. but the GCSDatasetdownloader should not be responsilbe for decompressing. This should happen in the Dataset class

datasetinsights/io/gcs.py
adason: +1

Also, we agree in the design meeting, the interface is 

```
download(self, *, url=None, local_path=None, bucket=None, key=None) 
```

datasetinsights/io/gcs.py
```diff
@@ -1,25 +1,34 @@
...
-        bucket = self.client.get_bucket(bucket_name)
-        blob = bucket.blob(object_key)
-
-        blob.download_to_filename(localfile)
+        """  # todo implement with file namedownload
+        if url:
+            bucket_name, key = self._parse(url)
+        return self._download_files(bucket_name, key, local_path)
 
     def upload(self, localfile, bucket_name, object_key):
```
adason:
```
upload(self, *, local_path=None, url=None, bucket=None, key=None) 
```

datasetinsights/io/gcs.py
AdamPalmarUnity: Also include input validation checks as bucket_name + key is mutually exclusive with url. It should be clear what takes precedence.

datasetinsights/io/gcs.py
```diff
@@ -29,6 +38,78 @@ def upload(self, localfile, bucket_name, object_key):
...
+
+    def _parse(self, url):
+        gcs_prefix = "gs://"
+        if not url.startswith(gcs_prefix):
+            raise ValueError(
+                f"Specified destination prefix: {url} does not start "
+                f"with {gcs_prefix}"
+            )
+        url = url[len(gcs_prefix) :]
+        idx = url.index("/")
```
AdamPalmarUnity: This will throw a  `ValueError: substring not found` for the url gs://bucket

datasetinsights/io/gcs.py
```diff
@@ -29,6 +38,78 @@ def upload(self, localfile, bucket_name, object_key):
 
         blob.upload_from_filename(localfile)
 
+    def _download_files(self, bucket_name, key, local_path):
```
AdamPalmarUnity: This method is not very readable can you refactor it so its more clear what its doing?

This part specifically 
```
     
       
       for blob in blobs:
            self._download_file(key, local_path, blob)
            if blob.name == key:
                is_file = True
                local_file_path = self._local_filename(key, blob, local_path)
            if not valid_key:
                valid_key = True
        if is_file:
            return local_file_path`
```

datasetinsights/io/gcs.py
```diff
@@ -29,6 +38,78 @@ def upload(self, localfile, bucket_name, object_key):
...
+            if not valid_key:
+                valid_key = True
+        if is_file:
+            return local_file_path
+        elif not valid_key:  # TODO sanjay check direct api
+            raise ValueError(f"Specified key: {key} does not exists ")
+
+        return local_path
+
+    def _download_file(self, key, local_path, blob):
```
AdamPalmarUnity: What is the key used for? Would also add some helper private methods to make it more readable.

datasetinsights/io/gcs.py
```diff
@@ -1,25 +1,34 @@
...
+
 
 class GCSClient:
     def __init__(self, **kwargs):
         """ Initialize a client to google cloud storage (GCS).
         """
         self.client = Client(**kwargs)
 
-    def download(self, bucket_name, object_key, localfile):
+    def download(
```
AdamPalmarUnity: Add tests for the two cases. Downloading a folder vs a file,

datasetinsights/io/gcs.py
```diff
@@ -38,32 +56,32 @@ def upload(self, localfile, bucket_name, object_key):
...
-            if blob.name == key:
-                is_file = True
-                local_file_path = self._local_filename(key, blob, local_path)
-            if not valid_key:
-                valid_key = True
-        if is_file:
-            return local_file_path
-        elif not valid_key:  # TODO sanjay check direct api
-            raise ValueError(f"Specified key: {key} does not exists ")
+            local_file_path = blob.name.replace(key, local_path)
```
AdamPalmarUnity, line 65: I would refactor code duplication into a private method.
```
        local_file_path = blob.name.replace(key_suffix, local_path)
        dst_dir = local_file_path.replace("/" + basename(local_file_path), "")
        self._download_blob(blob, dst_dir, key, local_file_path)
        self._checksum(blob, local_file_path)
```

AdamPalmarUnity (commented):
For the future add a better commit message that describes what is being commited/changed/updated.

datasetinsights/io/gcs.py
```diff
@@ -38,32 +56,32 @@ def upload(self, localfile, bucket_name, object_key):
 
         blob.upload_from_filename(localfile)
 
-    def _download_files(self, bucket_name, key, local_path):
-        bucket = self.client.get_bucket(bucket_name)
+    def _download_folder(self, bucket, key, local_path):
         blobs = bucket.list_blobs(prefix=key)
```
AdamPalmarUnity, line 63: The method bucket.list_blobs returns a iterator. It might be good to specify this in the variable name. This could be a variable name like  blob_iterator.

datasetinsights/io/gcs.py
```diff
@@ -24,11 +24,29 @@ def __init__(self, **kwargs):
...
+        """
         if url:
             bucket_name, key = self._parse(url)
-        return self._download_files(bucket_name, key, local_path)
+
+        bucket = self.client.get_bucket(bucket_name)
+        if self._is_file(bucket, key):
+            return self._download_file(bucket, key, local_path)
+        else:
+            return self._download_folder(bucket, key, local_path)
```
AdamPalmarUnity: Why is the local_path being returned here?

datasetinsights/io/gcs.py
```diff
@@ -24,11 +24,29 @@ def __init__(self, **kwargs):
...
+                GCS the dataset should be downloaded from.
+                The expected source-uri follows these patterns
+                gs://bucket/folder or gs://bucket/folder/data.zip
+
+            output: This is the path to the directory
+                where the download will store the dataset.
+                Examples:
+            >>> url = "gs://bucket/folder or gs://bucket/folder/data.zip"
+            >>> local_path = "/tmp/folder"
+            >>> output ="/tmp/folder or /tmp/folder/data.zip"
```
AdamPalmarUnity: There is no parameter in this method that is called output.

datasetinsights/io/gcs.py
```diff
@@ -38,32 +56,32 @@ def upload(self, localfile, bucket_name, object_key):
...
-            raise ValueError(f"Specified key: {key} does not exists ")
+            local_file_path = blob.name.replace(key, local_path)
+            dst_dir = local_file_path.replace("/" + basename(local_file_path), "")
+            self._download_blob(blob, dst_dir, key, local_file_path)
+            self._checksum(blob, local_file_path)
+            if not any_blob:
+                any_blob = True
+
+        if not any_blob:
+            raise ValueError(f"Specified key: {key} does not have file to download ")
```
AdamPalmarUnity: I don't think this should be a ValueError. It would probably fit more if this was a custom exception or FileNotFoundError although this seems to indicate that the file is located locally

datasetinsights/io/gcs.py
```diff
@@ -29,6 +56,77 @@ def upload(self, localfile, bucket_name, object_key):
...
 
+    def _download_folder(self, bucket, key, local_path):
+        blobs = bucket.list_blobs(prefix=key)
+        any_blob = False
+        for blob in blobs:
+            local_file_path = blob.name.replace(key, local_path)
+            dst_dir = local_file_path.replace("/" + basename(local_file_path), "")
+            self._download_blob(blob, dst_dir, key, local_file_path)
+            self._checksum(blob, local_file_path)
+            if not any_blob:
```
AdamPalmarUnity: Is there a need for this None check?

datasetinsights/io/gcs.py
```diff
@@ -1,25 +1,52 @@
...
 
 from google.cloud.storage import Client
 
+from datasetinsights.io.download import validate_checksum
+from datasetinsights.io.exceptions import ChecksumError, DownloadError
+
 logger = logging.getLogger(__name__)
 
+MD5 = "MD5"
+REPAD = "=="
```
BlairLee: Why do we need this constant?

datasetinsights/io/download.py
sanjayuconn: GCS encode MD5 hash of the data; encoded using base64

Ref: https://cloud.google.com/storage/docs/json_api/v1/objects

datasetinsights/io/gcs.py
sanjayuconn: WIP.

datasetinsights/io/gcs.py
sanjayuconn: WIP

datasetinsights/io/gcs.py
```diff
@@ -1,25 +1,57 @@
...
+    """
+
     def __init__(self, **kwargs):
         """ Initialize a client to google cloud storage (GCS).
         """
         self.client = Client(**kwargs)
 
-    def download(self, bucket_name, object_key, localfile):
-        """ Download a single object from GCS
+    def download(self, *, url=None, local_path=None, bucket=None, key=None):
```
AdamPalmarUnity, line 31: Validate that the key and bucket are not None if url is not provided.

datasetinsights/io/gcs.py
```diff
@@ -29,16 +61,91 @@ def upload(self, localfile, bucket_name, object_key):
...
+        self._download_validate(blob, key, local_file_path)
+
+    def _download_validate(self, blob, key, local_file_path):
+        """ download file and validate checksum
+        """
+        dst_dir = local_file_path.replace("/" + basename(local_file_path), "")
+        self._download_blob(blob, dst_dir, key, local_file_path)
+        self._checksum(blob, local_file_path)
+
+    def _download_blob(self, blob, dst_dir, key, local_file_path):
```
AdamPalmarUnity: What is the parameter key used for? It seems the information needed is in the blob object already.

tests/test_gcs.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from datasetinsights.io.gcs import GCSClient, gcs_bucket_and_path
+from datasetinsights.io.gcs import GCSClient, parse_gcs_location
 
 
 def test_gcs_client_warpper():
```
AdamPalmarUnity: Typo

AdamPalmarUnity: Also this test is doing multiple test cases. It should be refactored into multiple tests, upload/download single fie/download folder.

tests/test_gcs.py
```diff
@@ -22,24 +22,28 @@ def test_gcs_client_warpper():
...
+        client.download(
+            local_path=localfile, bucket=bucket_name, key=object_key
+        )
         mocked_gcs_client.get_bucket.assert_called_with(bucket_name)
-        mocked_bucket.blob.assert_called_with(object_key)
-        mocked_blob.download_to_filename.assert_called_with(localfile)
+        mocked_bucket.list_blobs.assert_called_with(prefix=object_key)
+        # # mocked_bucket.blob.assert_called_with(object_key)
+        # mocked_blob.download_to_filename.assert_called_with(localfile)
+        # todo file download test and folder download test # mock anyblob
```
AdamPalmarUnity: Put these test in a seperate unittest.

datasetinsights/io/gcs.py
sanjayuconn: If bucket is non GCS API will throw error

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+        path = url[(idx + 1) :]
+
+        return bucket, path
+
+    def upload(
+        self, *, local_path=None, bucket=None, key=None, url=None, pattern="*"
+    ):
+        """ Upload a file or list of files from directory to GCS
+
+            Args:
```
adason, line 153: Please fix docstring indentation.

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+            self._upload_file(local_path=local_path, bucket=bucket_obj, key=key)
+
+    def _upload_file(self, local_path=None, bucket=None, key=None):
+        """ Upload a single object to GCS
+        """
+        blob = bucket.blob(key)
+        try:
+            logger.info(f"Uploading from {local_path} to {key}.")
+            blob.upload_from_filename(local_path)
+        except UploadError as e:
```
adason: When is `UploadError` raised? Will it be raised by `blob.upload_from_filename(...)`?

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+    def _MD5_hex(self, checksum):
+        """fix the missing padding if requires and converts into hex"""
+        missing_padding = len(checksum) % 4
+        if missing_padding != 0:
+            checksum += "=" * (4 - missing_padding)
+        return base64.b64decode(checksum).hex()
+
+    def _parse(self, url):
+        """Split an GCS-prefixed URL into bucket and path."""
+        gcs_prefix = "gs://"
```
adason: `gcs_prefix` and `key_separator` should be class CONSTANT variables.

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+        key_separator = "/"
+        if not url.startswith(gcs_prefix):
```
adason: [nit] Might be easier using `regex` to match strings.

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+                raise e
+
+    def _is_file(self, bucket, key):
+        """given key is file or directory"""
+        blob = bucket.get_blob(key)
+        if blob:
+            return blob.name == key
+        return False
+
+    def _MD5_hex(self, checksum):
```
adason: PEP8. Use method name `_md5_hex` in favor of `_MD5_hex`. 
> Use the function naming rules: lowercase with words separated by underscores as necessary to improve readability.

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+                )
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Delete the downloaded files.")
+                os.remove(dst_file_name)
+                raise e
+
+    def _is_file(self, bucket, key):
+        """given key is file or directory"""
+        blob = bucket.get_blob(key)
+        if blob:
```
adason: [nit] Can be simplified to 1-line such as `return blob and blob.name == key`

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+                validate_checksum(
+                    dst_file_name, expected_checksum_hex, algorithm="MD5"
+                )
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Delete the downloaded files.")
+                os.remove(dst_file_name)
+                raise e
+
+    def _is_file(self, bucket, key):
+        """given key is file or directory"""
```
adason: [nit] Docstring should be improved. I suggest "Check if the key is a file or directory."

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+                           complete
+        """
+        dst_dir = local_file_path.replace("/" + basename(local_file_path), "")
+        key = blob.name
+        if not isdir(dst_dir):
+            makedirs(dst_dir)
+        try:
+            logger.info(f"Downloading from {key} to {local_file_path}.")
+            blob.download_to_filename(local_file_path)
+        except DownloadError as e:
```
adason: When is `DownloadError` raised? Will it be raised by blob.download_from_filename(...)?

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+            logger.info(
```
adason: Use `logger.exception`

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+        """
+        expected_checksum = blob.md5_hash
+        if expected_checksum:
+            expected_checksum_hex = self._MD5_hex(expected_checksum)
+            try:
+                validate_checksum(
+                    dst_file_name, expected_checksum_hex, algorithm="MD5"
+                )
+            except ChecksumError as e:
+                logger.info("Checksum mismatch. Delete the downloaded files.")
```
adason: Use `logger.exception`

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+            logger.info(f"Downloading from {key} to {local_file_path}.")
+            blob.download_to_filename(local_file_path)
+        except DownloadError as e:
+            logger.info(
+                f"The request download from {key} -> {local_file_path} can't "
+                f"be completed."
+            )
+            raise e
+
+    def _checksum(self, blob, dst_file_name):
```
adason: [nit] replace `dest_file_name`  with `filename`

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+        bucket_obj = self.client.get_bucket(bucket)
+        if self._is_file(bucket_obj, key):
+            self._download_file(bucket_obj, key, local_path)
+        else:
+            self._download_folder(bucket_obj, key, local_path)
 
+    def _download_folder(self, bucket, key, local_path):
+        """ download all files from directory
+        """
+        blobs = bucket.list_blobs(prefix=key)
```
adason, line 63: Will `bucket.list_blobs` recursively go through the GCS storage hierarchy? Will `_downliad_folder` download all files recursively?

datasetinsights/io/gcs.py
```diff
@@ -1,103 +1,227 @@
...
+        for blob in blobs:
+            local_file_path = blob.name.replace(key, local_path)
+            self._download_validate(blob, local_file_path)
```
adason, line 66: [nit] Why not reuse `_download_file` like what you have in `_upload_file`?

datasetinsights/datasets/groceries_real.py
```diff
@@ -324,17 +324,16 @@ def download(self):
         """Download dataset from GCS
         """
         cloud_path = f"gs://{const.GCS_BUCKET}/{self.GCS_PATH}"
+        client = GCSClient()
```
adason, line 361: Please remove download code in GroceriesReal. `Dataset` class will not download files going forward, and GroceriesReal is a public dataset that can be downloaded through http directly.

datasetinsights/commands/download.py
```diff
@@ -39,10 +39,7 @@ def convert(self, value, param, ctx):
         return value
 
 
-@click.command(
-    help="Download dataset to localhost from known locations.",
-    context_settings=const.CONTEXT_SETTINGS,
-)
+@click.command(context_settings=const.CONTEXT_SETTINGS,)
```
adason: Pull from master and add only GCSDownloader related instructions.

adason (changes requested):

datasetinsights/io/gcs.py
AdamPalmarUnity: There is no folder hierarchy on GCS, the are only keys file files. I think the prefix catches all that come after that prefix which is similar to a folder structure.

datasetinsights/io/gcs.py
AdamPalmarUnity: +1

tests/test_gcs.py
```diff
@@ -1,45 +1,390 @@
...
+local_path = "path/to/local"
+md5_hash = "abc=="
+md5_hash_hex = "12345"
+file_name = "/data.zip"
 
-def test_gcs_client_warpper():
-    bucket_name = "fake_bucket"
+
+@patch("datasetinsights.io.gcs.isdir")
+def test_gcs_client_upload_file(mock_isdir):
```
AdamPalmarUnity: This test is testing two different behaviors it seems

tests/test_gcs.py
```diff
@@ -1,45 +1,390 @@
...
+        mocked_blob.upload_from_filename.assert_called_with(localfile)
+
+        client.upload(local_path=localfile, url=url)
+        mocked_gcs_client.get_bucket.assert_called_with(bucket_name)
+        assert mocked_blob.upload_from_filename.call_count == 2
 
+
+@patch("datasetinsights.io.gcs.Path.glob")
+@patch("datasetinsights.io.gcs.isdir")
+def test_gcs_client_upload_folder(mock_isdir, mock_glob):
```
AdamPalmarUnity: Same as above

tests/test_gcs.py
```diff
@@ -1,45 +1,390 @@
...
+        )
+
+        client.upload(local_path=local_path, url=url)
+        mocked_gcs_client.get_bucket.assert_called_with(bucket_name)
+        assert client._upload_file.call_count == 2
+
+
+@patch("datasetinsights.io.gcs.isdir")
+@patch("datasetinsights.io.gcs.validate_checksum")
+def test_gcs_client_download_file(mock_checksum, mock_isdir):
```
AdamPalmarUnity: This test is testing two different behaviors it seems

tests/test_gcs.py
```diff
@@ -1,45 +1,390 @@
...
+
+        mock_checksum.assert_called_with(
+            local_file_path, md5_hash_hex, algorithm="MD5"
+        )
+        assert mock_checksum.call_count == 2
+
+
+@patch("datasetinsights.io.gcs.isdir")
+@patch("datasetinsights.io.gcs.validate_checksum")
+def test_gcs_client_download_folder(mock_checksum, mock_isdir):
```
AdamPalmarUnity: This test is testing two different behaviors it seems

AdamPalmarUnity (commented):
A lot of the test should be broken into different tests as they are testing multiple things.

datasetinsights/datasets/groceries_real.py
```diff
@@ -358,17 +358,16 @@ def download(self):
...
         # download label file
-        label_zip = download_file_from_gcs(
-            cloud_path, self.root, self.LABEL_ZIP
-        )
+        label_zip = self.LABEL_ZIP
+        client.download(url=cloud_path, local_path=self.root)
         with zipfile.ZipFile(label_zip, "r") as zip_dir:
```
BlairLee, line 365: You can archive the downloaded file and use `shutil.unpack_archive`, check details in this [PR](https://github.com/Unity-Technologies/datasetinsights/pull/57/files#).

datasetinsights/io/gcs.py
sanjayuconn: Object keys for folder and file are different so it has different t way to preprocess the local_file_path, so we cannot couple both. Hence calling _download_validate  which is next step of _download_file.

datasetinsights/io/gcs.py
sanjayuconn: - Fixed

datasetinsights/io/gcs.py
sanjayuconn: blob.upload_from_filename does not raise any exception so remove exception handling.

datasetinsights/io/gcs.py
sanjayuconn: Fixed

datasetinsights/io/gcs.py
sanjayuconn: Fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/io/gcs.py
sanjayuconn: fixed

datasetinsights/datasets/groceries_real.py
sanjayuconn: You mean GoogleGroceriesReal? GroceriesReal does not have any download code.

datasetinsights/datasets/groceries_real.py
sanjayuconn: Dataset won't download anything going forward.

tests/test_gcs.py
sanjayuconn: fixed

tests/test_gcs.py
sanjayuconn: fixed

tests/test_gcs.py
sanjayuconn: fixed

tests/test_gcs.py
sanjayuconn: fixed

tests/test_gcs.py
AdamPalmarUnity: I don't see any difference in the test.

tests/test_gcs.py
AdamPalmarUnity: I don't see any difference in the test.

tests/test_gcs.py
AdamPalmarUnity: I don't see any difference in the test.

sanjayuconn: check now

tests/test_gcs.py
sanjayuconn: check now

AdamPalmarUnity (approved):

adason (approved):

PR #52, opened 2020-08-21
Update to download command help text.

AdamPalmarUnity:
# Peer Review Information

Adding help text to download command with examples

# Pull Request Check List


datasetinsights/commands/download.py
```diff
@@ -39,10 +39,7 @@ def convert(self, value, param, ctx):
         return value
 
 
-@click.command(
-    help="Download datasets to localhost from known locations.",
-    context_settings=const.CONTEXT_SETTINGS,
-)
+@click.command(context_settings=const.CONTEXT_SETTINGS,)
```
adason, line 42: Why extra comma here?

datasetinsights/commands/download.py
```diff
@@ -82,6 +79,24 @@ def convert(self, value, param, ctx):
...
 ):
+    """Download datasets to localhost from known locations.
+
+    The download command can support downloading from 3 sources
+    usim:// http(s):// gs://
+
+    Examples of --source-uri:
+
+    \b
+    Unity Simulation
```
adason: I think it will be better to provide full example instead of just source URI. For example, we should have something like

```
Download from Unity Simulation:

You can specify project_it, run_execution_id, access_token in source-uri

datasetinsights download --source-uri=usim://<access_token>@<project_id>/<run_execution_id> --output=$HOME/data

Alternatively, you can also override access_token such as

datasetinsights download --source-uri=usim://<project_id>/<run_execution_id> --output=$HOME/data --access-token=<access_token>



```

datasetinsights/commands/download.py
```diff
@@ -82,6 +79,24 @@ def convert(self, value, param, ctx):
...
+    usim://access_token@project_id/run_execution_id
+
+    Groceries
```
adason: We don't want to list all dataset here. Instead we should simply provide 

```
datasetinsights download --source-uri=http://url.to.file.zip --output=$HOME/data
```

datasetinsights/commands/download.py
```diff
@@ -82,6 +79,24 @@ def convert(self, value, param, ctx):
...
+    https://storage.googleapis.com/datasetinsights/data/groceries/v3.zip
+
+    Synthetic Sample
+    https://storage.googleapis.com/datasetinsights/data/synthetic/SynthDet.zip
+
```
adason, line 124: GCS command example?

adason (changes requested):

datasetinsights/commands/download.py
AdamPalmarUnity: This is added automatically by pre-commit

datasetinsights/commands/download.py
BlairLee: +1

adason (approved):

PR #57, opened 2020-08-22
Added Compression Module and Refactored Dataset Class

Saurav-D:
# Peer Review Information

- Added `compression.py` module
- Updated `synthetic.py` and `groceries_real.py` module
- Added tests for above modules
- Updated doc string for `SynDetection2D` class

# Pull Request Check List


datasetinsights/io/compression.py
```diff
@@ -0,0 +1,35 @@
...
+            extension = _get_file_extension_from_filepath(filepath)
+            shutil.unpack_archive(filepath, destination, extension)
+        except ValueError as e:
+            logger.debug(
+                f"Current file format is not supported for decompression."
+            )
+            raise e
+
+
+def _get_file_extension_from_filepath(filepath):
```
adason: Why do we prefer `filetype` module over python's build in standard library [mimetypes](https://docs.python.org/3.7/library/mimetypes.html)?

datasetinsights/io/compression.py
```diff
@@ -0,0 +1,35 @@
+import logging
+import shutil
+
+import filetype
+
+logger = logging.getLogger(__name__)
+
+
+class Compression:
```
adason: Why do we need a Compression class here? Why not just a python function?

datasetinsights/io/compression.py
```diff
@@ -0,0 +1,35 @@
...
+    @staticmethod
+    def decompress(filepath, destination):
+        try:
+            extension = _get_file_extension_from_filepath(filepath)
+            shutil.unpack_archive(filepath, destination, extension)
```
adason: According to their official [documentation](https://docs.python.org/3.7/library/shutil.html#shutil.unpack_archive), shutil.unpack_archive(...) can automatically detect file archive type. Why don't we just let it handle the task to guess compress type?

datasetinsights/io/compression.py
```diff
@@ -0,0 +1,35 @@
...
+import shutil
+
+import filetype
+
+logger = logging.getLogger(__name__)
+
+
+class Compression:
+    @staticmethod
+    def decompress(filepath, destination):
```
adason: I think it's better to rename this to `unpack_archive` that align with the name `shutil.unpack_archive` in python's standard library.

datasetinsights/io/__init__.py
```diff
@@ -9,3 +13,23 @@
     "KubeflowPipelineWriter",
     "create_downloader",
 ]
+
+
+def gzip_decompress(filepath, destination):
```
adason: Why is this method required? `shutil.unpack_archive` supports `gztar` which is `.tar.gz` file. I don't think it's very common to have `file.text.gz` file type.

datasetinsights/datasets/groceries_real.py
```diff
@@ -217,6 +237,15 @@ def _convert_to_bbox2d(single_bbox):
         )
         return canonical_bbox
 
+    @staticmethod
+    def is_groceries_real_dataset_files_present(dataset_directory):
```
adason: rename this to `is_dataset_files_present(...)`. Don't need to repeat groceries_real here.

datasetinsights/datasets/groceries_real.py
```diff
@@ -217,6 +237,15 @@ def _convert_to_bbox2d(single_bbox):
...
+        return (
+            os.path.isdir(dataset_directory)
+            and glob.glob(f"{dataset_directory}/**/*.JPG")
```
adason: I will make this condition more relaxed. See similar comments in `SynDetection2D`.

datasetinsights/datasets/synthetic.py
```diff
@@ -247,3 +289,11 @@ def exists(capture_file):
         keep_mask = catalog.filename.apply(exists)
 
         return catalog[keep_mask]
+
+    @staticmethod
+    def is_synthetic_dataset_files_present(dataset_directory):
```
adason: I will suggest we make this detection less strict. We should detect 
1. downloaded zip file exists
2. folder of the zip file unpacked exists
3. it's not an empty folder

We don't really want to check too much detail here. It's almost impossible to guess what types of files that will be included in the synthetic dataset. Also, binary files are not always required to use this dataset. Users might be interested in annotations only if they are simply want dataset statistics.

datasetinsights/datasets/synthetic.py
```diff
@@ -96,16 +99,32 @@ def read_bounding_box_2d(annotation, label_mappings=None):
...
-    Current public synthdet version are as follows:
-        v1: Subset SynthDet dataset containing around 5k images.
+    During the class instantiation, it would check whether the data files
+    such as annotations.json, images.png are present, if not it'll check
+    whether a compressed dataset file is present which contains the necessary
+    files, if not it'll raise an error.
+
+    The dataset contain the following structures:
+
+    * DatasetXXX-XXX-XXX
```
adason: We leave out this folder structure on purpose. This is actually something that might change in the future. Using the folder name `Dataset-UUID` is a temporary fix for some of the UnitySimulation issues that tries to write output files in different location. 

We should simply point to the Synthetic dataset schema documentation here.

datasetinsights/datasets/synthetic.py
```diff
@@ -96,16 +99,32 @@ def read_bounding_box_2d(annotation, label_mappings=None):
...
+
+    * RGBxXX-XXX-XXX
+    .
+    .
+    * RGBxXX-XXX-XXX
+        * rgb_XXX.jpg ... rgb_XXX.jpg
 
     Attributes:
-        root (str): root directory of the dataset
+        dataset_directory (str): root directory of the dataset
```
adason: I think this is potentially confusing with `data_path` args that user are expected to pass to the dataset class. I suggest we make this instance variable private and stick to the name `data_path`, such as `self._data_path`.

datasetinsights/datasets/groceries_real.py
```diff
@@ -103,7 +105,30 @@ def __init__(
             transforms: callable transformation
             version (str): version of GroceriesReal dataset
         """
-        self.data_path = data_path
+        # check if dataset files are present
```
adason: I suggest wrapping this logic to check if dataset exists into a single instance method or static method. This way we can have separate unittest for this method, it's easier to read, and it does not make the `__init__` method grow  in dramatically. 

Also, it will be nice to have a simple interface with the following logic, such as

def _prepare_dataset(data_path): # need a better name
    # check if dataset zip file exists. if not, raise DatasetNotFoundError
    # check if dataset is unzipped to a folder and the folder is not empty. if not, try to decompress it. if it can't decompress, raise some other error
    # everything works well, happily return

datasetinsights/datasets/synthetic.py
```diff
@@ -137,7 +156,30 @@ def __init__(
             random_seed (int): random seed used for splitting dataset into
                 train and val
         """
-        self.dataset_directory = data_path
+        # check if dataset files are present
+        if SynDetection2D.is_synthetic_dataset_files_present(data_path):
```
adason: similar to comments in GroceriesReal, we should simplify this to call an instance method or static method.

For Synthetic dataset, there are a few issues that we need to consider. Users should be able to use `SynDetection2D` for 1. a dataset downloaded by UnitySimulationDownloader. This will download files to `data_path` as the parent directory. 2. a dataset downloaded as a single zip file. This will downlad file to `data_path/file.zip`.

In case 1, we should directly set `self._data_path = data_path`. In the second case, you will want to go through similar checks as in GrocreiesReal. Once the files are decompressed, you will want to set `self._data_path = data_path/unzipped_folder`

adason (changes requested):

datasetinsights/io/compression.py
AdamPalmarUnity: The implementation behind it is rather naive. 

```
def _find_unpack_format(filename):
    for name, info in _UNPACK_FORMATS.items():
        for extension in info[0]:
            if filename.endswith(extension):
                return name
    return None
```

tests/datasets/test_synthetic.py
```diff
@@ -60,3 +62,20 @@ def test_get_split():
...
     pd.testing.assert_frame_equal(expected_val, actual_val)
+
+
+def test_is_synthetic_dataset_files_present_returns_true():
+    with tempfile.TemporaryDirectory() as tmp:
+        temp_dataset_dir = os.path.join(tmp, "Dataset893-908")
+        temp_rgb_dir = os.path.join(tmp, "RGB893-908")
+        os.mkdir(temp_dataset_dir)
+        os.mkdir(temp_rgb_dir)
+        open(os.path.join(temp_dataset_dir, "annotation.json"), "x")
```
AdamPalmarUnity: Would add the with statement here when creating the files.
with open() ...

tests/datasets/test_groceriesreal.py
```diff
@@ -0,0 +1,20 @@
...
+import tempfile
+
+from datasetinsights.datasets.groceries_real import GroceriesReal
+
+
+def test_is_groceries_real_dataset_files_present_returns_true():
+    with tempfile.TemporaryDirectory() as tmp:
+        temp_images_dir = os.path.join(tmp, "images")
+        os.mkdir(temp_images_dir)
+        open(os.path.join(temp_images_dir, "IMG_4185.JPG"), "x")
```
AdamPalmarUnity: Would add with statement here.

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: +1
But whether or not this is refactored into a method the important part to call out is that the construction of the dataset object is tested.
Secondly if there needs to be a comment above come code block indicates that there should be a method/function that makes it easier to read although, its a minor improvement i think.

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: 1+

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: def _preprocess_dataset could be the name maybe? At least that is very generic

tests/datasets/test_synthetic.py
BlairLee: +1

datasetinsights/io/compression.py
adason: I actually think this implementation in standard library is good enough. I would prefer not have to maintain code such as functions defined in `_get_file_extension_from_filepath` if this is already handled by standard library, unless this method tries to do something that is very different, or some side effects are required but not implemented in standard library...

datasetinsights/io/compression.py
Saurav-D: Mimetypes does not work for most of the file types. It just returns none.

datasetinsights/io/compression.py
Saurav-D: We can make it a python function as well.

datasetinsights/io/compression.py
Saurav-D: `_get_file_extension_from_filepath ` detect file extension by reading the file header/data instead of file name which is done in standard library. Keeping the current implementation will also help us when downloading files from HTTP when we dont know the file name and file type so we can just download it without giving any extension to it and `_get_file_extension_from_filepath` will automatically detect filetype and do the necessary operation.

datasetinsights/datasets/groceries_real.py
Saurav-D: will do.

datasetinsights/datasets/synthetic.py
Saurav-D: Is there any link to the schema documentation? Or it is yet to be done?

datasetinsights/datasets/groceries_real.py
Saurav-D: So just to be clear only way we check if dataset files are present (either in data_path or after unpacking dataset archive file) is to: 
1. If data_path directory exists
and
2. If the directory is not empty

datasetinsights/datasets/synthetic.py
Saurav-D: But how do we check for empty folder? because we would unpack dataset archive file in the same directory and what if the archive file is empty then the dataset archive file would still be there and give False for empty directory check.

datasetinsights/io/compression.py
Saurav-D: Wouldn't that create ambiguity?

datasetinsights/datasets/synthetic.py
adason: https://datasetinsights.readthedocs.io/en/latest/Synthetic_Dataset_Schema.html

datasetinsights/io/compression.py
adason: This is the [list](https://www.iana.org/assignments/media-types/media-types.xhtml 
) of supported file types, according to the official [documentation](https://docs.python.org/3.7/library/mimetypes.html#mimetypes.guess_type). It actually covers more than what's covered by `filetype` package. Why is `filetype` a better replacement for the standard library?

datasetinsights/io/compression.py
adason: Why do we have to change file extension after the file was downloaded? Shouldn't we simply take the file as is? It's very unlikely we will encounter a dataset that have incorrect filename extension. If this is the issue in GroceriesReal and SynthDet sample dataset that we have control, we should make the filename extension correct when the file is downloaded.

datasetinsights/datasets/groceries_real.py
adason:
```
from pathlib import Path 

def exists():
    p = Path(data_path) / "folder_name"
    return p.exits() and len(p.glob("**/*")) > 0
```

datasetinsights/datasets/synthetic.py
adason:
```
from pathlib import Path 

def exists():
    p = Path(data_path) / "folder_name"
    return p.exits() and any(p.glob("**/*"))
```

datasetinsights/io/__init__.py
```diff
@@ -9,3 +13,23 @@
...
+    """ Decompress a gzip file to the destination.
```
adason: Did you build the documentation to make sure this module is documentation is included? Are the docstrings parsed properly?

datasetinsights/datasets/synthetic.py
Saurav-D: Right now, pulbic synthdet dataset as well as groceries real dataset files are zipped into single zip file. And HTTP downloader unzips all files directly to `data_path`. Why do we need to unzip them into `unzip_folder` because then dataset class need to know unzip folder for each dataset, on the other hand Dataset class just look in `data_path` for filed and don't need to know anything about `unzipped_folder`.

datasetinsights/io/__init__.py
Saurav-D: We can remove this, as it is not needed for now.

datasetinsights/io/__init__.py
Saurav-D: Removed this as it is not necessary right now.

datasetinsights/datasets/groceries_real.py
```diff
@@ -217,6 +214,39 @@ def _convert_to_bbox2d(single_bbox):
         )
         return canonical_bbox
 
+    @staticmethod
+    def _preprocess_dataset(data_path):
```
AdamPalmarUnity: If i understand this correctly this method will __always__ try to unzip the "dataset" file. Is it possible to only unzip if needed? Furthermore this method and the one in SynDetection2D puts the dataset in a hidden path either synthetic or groceries_real. This mean that two calls to train or evaluate will give different results if the unzip is changed to not happen every time. The first that unzips will point to the unzipped folder, the other will point to the data_path given.

datasetinsights/datasets/groceries_real.py
```diff
@@ -217,6 +214,39 @@ def _convert_to_bbox2d(single_bbox):
...
+            # check if necessary files are present after decompression
+            if GroceriesReal.is_dataset_files_present(unzip_folder):
+                return unzip_folder
+            else:
+                raise DatasetNotFoundError(
+                    f"No dataset file(s) present at path" f":{unzip_folder}"
+                )
+
+        # check if dataset files are present
+        if GroceriesReal.is_dataset_files_present(data_path):
```
AdamPalmarUnity, line 230: Should this method call be before the check to see if the compressed dataset file is present?

datasetinsights/datasets/groceries_real.py
AdamPalmarUnity: Also can you add some tests for the construction of the GroceriesReal and SynDetection2D with the cases where decompressing happens and for calls where DatasetNotFoundError is raised.

AdamPalmarUnity (dismissed):
Add tests for instantiation of Dataset classes. Make sure the correct errors are raised and so on.

datasetinsights/datasets/groceries_real.py
Saurav-D: We can unzip directly in data_path instead of unzip_folder.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -51,3 +62,74 @@ def download(self, source_uri, output, checksum_file=None, **kwargs):
                 logger.info("Checksum mismatch. Deleting the downloaded file.")
                 os.remove(dataset_path)
                 raise e
+
+
+def download_dataset_from_http_url(source_uri, output):
```
adason: Can this method reuse `datasetinsights.io.download.download_file`? This method seems to be a duplicate. You might want to refactor `download_file` to have the following signature:

```
# in datasetinsights.io.download
def http_download(source_uri, dest_path, filename=None, use_cache=True):
    # make http request to source_uri
    try:
    except:
    else:
        if not filename:
            filename = parse_filename(http.response)
        output = os.path.join(dest_path, filename)
        with open(output, "wb") as f:
                f.write(response.content)   
```

Also, you should make this private static method for `HTTPDatasetDownloader` that captures logic not in `http_download` (e.g. capture `DownloadError` and provide logging messages, re-raise, validation, ...)

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -51,3 +62,74 @@ def download(self, source_uri, output, checksum_file=None, **kwargs):
...
+    cd = response.headers.get("content-disposition")
+    if not cd:
+        return None
+    file_name = re.findall("filename=(.+)", cd)
+    if len(file_name) == 0:
+        return None
+    return file_name[0]
+
+
+def get_file_name_from_uri(uri):
```
adason: URL is not always reliable to parse filenames. It might have http redirect which hides details of files it contains. It should be sufficient to get filename from http responses.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -51,3 +62,74 @@ def download(self, source_uri, output, checksum_file=None, **kwargs):
...
+                file_name = get_file_name_from_uri(source_uri)
+
+            dataset_path = output / file_name
+
+            with open(dataset_path, "wb") as f:
+                f.write(response.content)
+    return dataset_path
+
+
+def get_filename_from_response(response):
```
adason: Consider making this private method.

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -51,3 +62,74 @@ def download(self, source_uri, output, checksum_file=None, **kwargs):
...
+    """ Gets filename from requests response object
+
+        Args:
```
adason: Have you checked if the docstrings are parsed correctly in the documentation?

datasetinsights/io/downloader/http_downloader.py
```diff
@@ -1,13 +1,26 @@
...
-    download_file,
+    TimeoutHTTPAdapter,
     get_checksum_from_file,
     validate_checksum,
 )
 from datasetinsights.io.downloader.base import DatasetDownloader
-from datasetinsights.io.exceptions import ChecksumError
+from datasetinsights.io.exceptions import ChecksumError, DownloadError
+
+# number of workers for ThreadPoolExecutor. This is the default value
```
adason: Why duplicate these constants that exist in `datasetinsights.io.download`

datasetinsights/datasets/synthetic.py
```diff
@@ -247,3 +250,43 @@ def exists(capture_file):
...
+
+        unarchived_path = Path(data_path) / SynDetection2D.SUBFOLDER
+        shutil.unpack_archive(
+            filename=archive_file, extract_dir=unarchived_path
+        )
+
+        return unarchived_path
+
+    @staticmethod
+    def is_dataset_files_present(data_directory):
```
BlairLee: It's better to keep the name consistent. E.g. `data_directory ` to `data_path`.

datasetinsights/io/downloader/http_downloader.py
Saurav-D: Not every time HTTP response contains content-disposition header.

adason (dismissed):

adason (approved):

PR #63, opened 2020-08-26
Fix Http Download Cache

adason:
# Peer Review Information

This change fixes a few issue of the current http download method. Previously, if `dest_path` as a folder already exist for estimator checkpoints, it will skip download. This change completed removed "cache" behavior and we should tackle dataset caching in a separate review. 

This change also improve evaluation notebook to make the interactive widget 

# Pull Request Check List

### DIFF
--- a/datasetinsights/io/checkpoint.py
+++ b/datasetinsights/io/checkpoint.py
@@ -247,8 +247,7 @@ def load_from_http(estimator, url):
 
     """
     with tempfile.TemporaryDirectory() as temp_dir:
-        path = os.path.join(temp_dir, "estimator_checkpoint")
-        logger.debug(f"Downloading estimator from {url} to {path}")
-        download_file(source_uri=url, dest_path=path)
+        logger.debug(f"Downloading estimator from {url} to {temp_dir}")
+        path = download_file(source_uri=url, dest_path=temp_dir)
         logger.debug(f"Loading estimator from {path}")
         estimator.load(path)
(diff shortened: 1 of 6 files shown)


datasetinsights/io/download.py
```diff
@@ -29,25 +29,17 @@ def send(self, request, **kwargs):
...
         file_name (str): file name of the file to be downloaded
-        use_cache (bool): use_cache (bool): use cache instead of
-        re-download if file exists
 
     Returns:
         String of destination path.
     """
-    dest_path = Path(dest_path)
-    if dest_path.exists() and use_cache:
-        return dest_path
```
BlairLee, line 49: I'm about to fix that! Nice work!

datasetinsights/io/checkpoint.py
```diff
@@ -247,8 +247,7 @@ def load_from_http(estimator, url):
 
     """
     with tempfile.TemporaryDirectory() as temp_dir:
-        path = os.path.join(temp_dir, "estimator_checkpoint")
-        logger.debug(f"Downloading estimator from {url} to {path}")
-        download_file(source_uri=url, dest_path=path)
```
BlairLee, line 252: Why would this implementation raise `IsADirectoryError` exception?

adason: Peviously implementation does not call `download_file` correctly as `<temp_dir>/estimator_checkpoint` is somehow downloaded as a directory. This will fail when estimator tries to load a checkpoint file which points to a directory.

Saurav-D (approved):

PR #67, opened 2020-08-27
Adding sphinx-click as a dependency and added docstrings.

AdamPalmarUnity:
# Peer Review Information

Adding sphinx-click as a dependency and added docstrings.
Also modified datasetinsights.commands.rst to show click documentation.

# Pull Request Check List


datasetinsights/io/downloader/base.py
```diff
@@ -63,4 +65,14 @@ def __init_subclass__(cls, protocol=None, **kwargs):
 
     @abstractmethod
     def download(self, source_uri, output, **kwargs):
+        """
```
adason: PEP257 suggests docstring should include titles.

datasetinsights/io/downloader/base.py
```diff
@@ -63,4 +65,14 @@ def __init_subclass__(cls, protocol=None, **kwargs):
...
+
+        Args:
+            source_uri: URI that points to the dataset that should be downloaded
+            output: path to local folder where the dataset should be stored
+            **kwargs:
+
+        Returns:
```
adason: no return value in this method

datasetinsights/io/downloader/base.py
```diff
@@ -48,6 +48,8 @@ def create_downloader(source_uri, **kwargs):
 
 
 class DatasetDownloader(ABC):
+    """This is the base class for all dataset downloaders"""
```
adason: It would be valuable to add a little context in the docstring here. We should describe required instance methods for this abstract interface, required parameters, ....Unless this can be automatically generated by sphinx given that this class is a subclass of `ABC` and we have a few `@abstractmethod` here.

adason (approved):

PR #70, opened 2020-08-27
Fix bug in bbox2d plots

BlairLee:
# Peer Review Information

Previously, for the label image rendering, if the label image exceeds the image border, it would raise a ValueError, like: `ValueError: could not broadcast input array from shape (93,1395,3) into shape (93,1279,3)`. Add crop function in `_add_single_bbox_on_image.py`.

# Pull Request Check List


datasetinsights/stats/visualization/bbox2d_plot.py
```diff
@@ -159,4 +159,21 @@ def _add_single_bbox_on_image(
 
         _cv2.rectangle(image, rec_left_top, rec_right_bottom, color, -1)
 
-        image[label_top:label_bottom, label_left:label_right, :] = label_image
+        label_top = max(0, label_top)
```
adason: Maybe line 162-179 can be wrapped in a function?

datasetinsights/stats/visualization/bbox2d_plot.py
```diff
@@ -158,5 +161,44 @@ def _add_single_bbox_on_image(
         rec_right_bottom = (rectangle_right, rectangle_bottom)
 
         _cv2.rectangle(image, rec_left_top, rec_right_bottom, color, -1)
+        _add_label_on_image(
```
adason: This method does not "add label" instead it tries to do label cropping. Since you have a method called `_add_label_on_image`, why not just move everything from line 135 to 163 to this method? I would expect to see this 
```
def _add_single_bbox_on_image()
  ...
  if label:
     _add_label_on_image(...)

def _add_label_on_image(...):
   ...
   _cv2.rectangle(...)
   _ fix_labe_at_image_edge(...)

def _fix_label_at_image_edge(...)
   ...
```

BlairLee: Fixed.

adason (approved):

PR #73, opened 2020-08-28
Bug fix torch distributed

AdamPalmarUnity:
# Peer Review Information

Refactor to _init_distributed_mode initialization now covers the gpu 0 and rank 0.
Also renamed the main.py module to '__main__.py' in order to be able to invoke it through python -m torch.distributed.launch
Change the kubeflow pipeline to have the flag --use_env such that the local_rank is passed to the program as an environment variable and not a argument.  

# Pull Request Check List


datasetinsights/datasets/synthetic.py
```diff
@@ -264,8 +264,8 @@ def _preprocess_dataset(data_path):
 
         # This is for dataset generated by unity simulation.
         # In this case, all data are downloaded directly in the data_path
-        if SynDetection2D.is_dataset_files_present(data_path):
-            return data_path
+        # if SynDetection2D.is_dataset_files_present(data_path):
```
AdamPalmarUnity: This breaks the case where a synthetic dataset is downloaded from Unity Simulation

datasetinsights/datasets/groceries_real.py
```diff
@@ -227,22 +228,31 @@ def _preprocess_dataset(data_path, version):
...
+                unarchived_path = Path(data_path) / GroceriesReal.SUBFOLDER
+                if not GroceriesReal.is_dataset_files_present(unarchived_path):
+                    shutil.unpack_archive(
+                        filename=archive_file, extract_dir=unarchived_path
+                    )
+
+                return unarchived_path
+            finally:
+                os.close(file_descriptor)
+        elif GroceriesReal.is_dataset_files_present(data_path):
```
adason, line 247: This test is made the first check in `_process_dataset` so that we don't have to check if archive_file exists. After-all, having archive file os a backup when the dataset is not unarchived.

datasetinsights/datasets/groceries_real.py
```diff
@@ -227,22 +228,31 @@ def _preprocess_dataset(data_path, version):
...
 
         archive_file = Path(data_path) / GroceriesReal.ARCHIVE_FILES[version]
-        if not archive_file.exists():
+        if archive_file.exists():
+            file_descriptor = os.open(archive_file, os.O_RDONLY)
+
+            try:
+                fcntl.flock(file_descriptor, fcntl.LOCK_EX)
+                unarchived_path = Path(data_path) / GroceriesReal.SUBFOLDER
+                if not GroceriesReal.is_dataset_files_present(unarchived_path):
```
adason, line 239: In general, nested if are harder to read and understand. Why do we prefer nested if over the current master which flattened these checks?

datasetinsights/datasets/synthetic.py
```diff
@@ -261,30 +262,33 @@ def _preprocess_dataset(data_path):
...
-        if SynDetection2D.is_dataset_files_present(data_path):
+        archive_file = Path(data_path) / SynDetection2D.ARCHIVE_FILE
+        if archive_file.exists():
+            file_descriptor = os.open(archive_file, os.O_RDONLY)
+
+            try:
+                fcntl.flock(file_descriptor, fcntl.LOCK_EX)
+
+                unarchived_path = Path(data_path) / SynDetection2D.SUBFOLDER
+                if not SynDetection2D.is_dataset_files_present(unarchived_path):
```
adason, line 273: Same as GroceriesReal comments.

datasetinsights/__main__.py
```diff
@@ -2,8 +2,8 @@
```
adason, line 2: Why rename file `main.py` to `__main__.py`?

AdamPalmarUnity: This is a fix to the error.
/usr/local/bin/python: can't find '__main__' module in 'datasetinsights'

adason (approved):

PR #74, opened 2020-08-31
Update cv depedency

AdamPalmarUnity:
# Peer Review Information

Add tests for create_estimator

# Pull Request Check List

### DIFF
--- a/tests/test_create_estimator.py
+++ b/tests/test_create_estimator.py
@@ -1,8 +1,7 @@
 import pytest
 from yacs.config import CfgNode as CN
 
-from datasetinsights.estimators import create_estimator
-from datasetinsights.estimators import FasterRCNN
+from datasetinsights.estimators import FasterRCNN, create_estimator
 
 
 @pytest.fixture
@@ -16,9 +15,6 @@ def config():
 
 def test_create_estimator_with_config_expect_faster_rcnn(config):
 
-    estimator = create_estimator(
-        name=config.estimator,
-        config=config,
-    )
+    estimator = create_estimator(name=config.estimator, config=config,)
 
     assert isinstance(estimator, FasterRCNN)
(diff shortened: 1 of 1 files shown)


tests/test_create_estimator.py
```diff
@@ -0,0 +1,20 @@
+import pytest
+from yacs.config import CfgNode as CN
+
+from datasetinsights.estimators import FasterRCNN, create_estimator
+
+
+@pytest.fixture
+def config():
```
adason, line 8: This test only tries to create estimator for `FasterRCNN`, as this is the only estimator defined in the yaml.

tests/test_create_estimator.py
```diff
@@ -0,0 +1,20 @@
...
+    """prepare config."""
+    with open("tests/configs/faster_rcnn_groceries_real_test.yaml") as f:
+        cfg = CN.load_cfg(f)
+
+    return cfg
+
+
+def test_create_estimator_with_config_expect_faster_rcnn(config):
```
adason, line 16: Test against a specific estimator create will become tedious,  as developers have to remember add a new test for each new estimator that was introduced. It's very likely that developers will forgot. Test whether `FasterRCNN` estimator can be successfully initialized should be the responsibility of tests specific to each estimator, such as `test_faster_rcnn.py` or `test_deeplab.py`

tests/test_create_estimator.py
```diff
@@ -0,0 +1,20 @@
...
+
+    estimator = create_estimator(name=config.estimator, config=config,)
```
adason, line 18: We should test if `_find_estimator` is working properly instead testing if a particular estimator can be initialized.

adason:
The title of this PR states `Update cv depedency` but it only modified python code without changing dependencies in either `pyproject.toml` or `Dockerfile`. Are those changes missing in this PR?

AdamPalmarUnity:
Yes the update to the depedency is not there yet.

PR #79, opened 2020-09-02
Improve synthetic image visualization in notebook

BlairLee:
# Peer Review Information

It would provide a default font size calculated based on the image resolution. Users do not have to “remember” what font size they should use.

### DIFF
--- a/datasetinsights/stats/visualization/plots.py
+++ b/datasetinsights/stats/visualization/plots.py
@@ -17,7 +17,8 @@ from datasetinsights.stats.visualization.bbox2d_plot import (
 
 logger = logging.getLogger(__name__)
 COLORS = list(ImageColor.colormap.values())
-
+FONT_SCALE = 35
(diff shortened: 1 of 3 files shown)


datasetinsights/stats/visualization/plots.py
```diff
@@ -149,10 +149,20 @@ def plot_bboxes(image, bboxes, label_mappings=None, colors=None):
         PIL Image: a PIL image with bounding boxes drawn.
     """
     np_image = np.array(image)
+    image_height, _, _ = np_image.shape
     for i, box in enumerate(bboxes):
         label = _process_label(box, label_mappings)
         color = colors[i] if colors else None
-        add_single_bbox_on_image(np_image, box, label, color)
+        font_size = image_height // 35
```
adason: Should make two magic values `35` and `250` constants.

adason (approved):

PR #94, opened 2020-09-15
Adding Pytest Config in pyproject.toml

Saurav-D:
# Peer Review Information

From pytest 6.0 we can write pytest configuration inside `pyproject.toml` file and therefore we don't need `pytest.ini` file in root. [Official Doc](https://docs.pytest.org/en/stable/customize.html#pyproject-toml)


adason (approved):
Excellent. One less file to keep track of.

PR #95, opened 2020-09-15
Update docstrings for metrics

BlairLee:
# Peer Review Information

Some of the metrics docstrings aren’t properly parsed. Update those improper docstrings.

# Pull Request Check List

### DIFF
--- a/datasetinsights/evaluation_metrics/average_precision_2d.py
+++ b/datasetinsights/evaluation_metrics/average_precision_2d.py
@@ -63,10 +63,10 @@ class AveragePrecision(EvaluationMetric):
 
         Args:
             mini_batch (list(list)): a list which contains batch_size of gt
-              bboxes and pred bboxes pair in each image. For example, if batch
-              size = 2, mini_batch looks like: [[gt_bboxes1, pred_bboxes1],
-              [gt_bboxes2, pred_bboxes2]] where gt_bboxes1, pred_bboxes1 contain
-              gt bboxes and pred bboxes in one image.
+                bboxes and pred bboxes pair in each image. For example, if batch
+                size = 2, mini_batch looks like: [[gt_bboxes1, pred_bboxes1],
+                [gt_bboxes2, pred_bboxes2]] where gt_bboxes1, pred_bboxes1
+                contain gt bboxes and pred bboxes in one image.
         """
         for bboxes in mini_batch:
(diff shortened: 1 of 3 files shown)


datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -1,6 +1,6 @@
 r"""Reference.
 
-https://github.com/rafaelpadilla/Object-Detection-Metrics#average-precision\
+https://github.com/rafaelpadilla/Object-Detection-Metrics#average-precision\.
```
adason: ??? Why add extra dot here?  Can you improve the docstring of this reference section?

datasetinsights/evaluation_metrics/average_precision_2d.py
```diff
@@ -60,12 +60,11 @@ def update(self, mini_batch):
...
 
         Args:
-            mini_batch (list(list)): a list which contains batch_size of
-            gt bboxes and pred bboxes pair in each image.
-            For example, if batch size = 2, mini_batch looks like:
-            [[gt_bboxes1, pred_bboxes1], [gt_bboxes2, pred_bboxes2]]
-            where gt_bboxes1, pred_bboxes1 contain gt bboxes and pred bboxes
-            in one image
+            mini_batch (list(list)): a list which contains batch_size of gt
+              bboxes and pred bboxes pair in each image. For example, if batch
```
adason: Why are these text indented by 2?

BlairLee: It would let docstring know that it belongs to one parameter. [Reference](https://google.github.io/styleguide/pyguide.html).

datasetinsights/evaluation_metrics/average_precision_2d.py
BlairLee: Will improve it.

datasetinsights/evaluation_metrics/average_precision_2d.py
adason: @BlairLee I was actually pointing out the indentation was not consistent. It should be 4 instead of 2.

BlairLee: Fixed.

Saurav-D (approved):

PR #97, opened 2020-09-17
Low hdd and memory size are causing large synthetic datasets to fail.

JonathanHUnity:
# Peer Review Information
Low hdd and memory size are causing large synthetic datasets to fail. Upping memory and hdd sizes to accomidate.

# Pull Request Check List

### DIFF
--- a/kubeflow/compiled/train_on_synthetic_dataset_unity_simulation.yaml
+++ b/kubeflow/compiled/train_on_synthetic_dataset_unity_simulation.yaml
@@ -2,7 +2,7 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: train-on-synthetic-dataset-unity-simulation-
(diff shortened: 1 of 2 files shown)


kubeflow/compiled/train_on_synthetic_dataset_unity_simulation.yaml
```diff
@@ -29,8 +29,8 @@ spec:
       - {name: CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE, value: /secret/gcp-credentials/user-gcp-sa.json}
       image: '{{inputs.parameters.docker}}'
       resources:
-        limits: {memory: 64Gi}
-        requests: {memory: 64Gi}
+        limits: {memory: 256Gi}
```
BlairLee, line 32: @adason  and I talked about that. We don't want to set a large default memory. Since some experiments don't need that big memory. Maybe we need to add some instructions about the issue in `Evaluation_Tutorial.md`, if you need to train a larger dataset (400K), we need to set the memory limit to 256Gi.

JonathanHUnity: Ok. I'll leave that to you guys then.

PR #105, opened 2020-09-30
Ground truth bounding boxes3d

StevenBorkman:
# Peer Review Information

Added 3D bounding box visualization to perception notebook. Made appropriate changes to perception notebook to re-enable 2D bounding box visualization.

# Pull Request Check List


datasetinsights/datasets/synthetic.py
```diff
@@ -70,6 +71,24 @@ def _get_split(*, split, catalog, train_percentage=0.9, random_seed=47):
             f"splits are: {VALID_SPLITS}"
         )
 
+def read_bounding_box_3d(annotation, label_mappings=None):
+    bboxes = []
+
```
JonathanHUnity, line 88: Needs docs

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -0,0 +1,146 @@
...
+        img_height (int): The height of the image in pixels.
+        img_width (int): The width of the image in pixels.
+
+    Returns:
+        numpy array: a one-dimensional array with two values (x and y) representing a point's
+        pixel coordinate in an image.
+    """
+
+    _pt = projection.dot(pt)
+    _pt /= _pt[2]
```
JonathanHUnity: What is this divison for, out of curiosity? Isn't it possible for the projected point to be at depth 0?

datasetinsights/stats/visualization/plots.py
```diff
@@ -127,6 +132,36 @@ def get_color(match):
     colors = [get_color(match) for _, match in match_results]
     return colors
 
+def plot_bboxes3d(image, bboxes, projection=None, colors=None):
+    """ Plot an image with 3D bounding boxes
+    
+    Currently only ground truth images are supported. If a list of colors is not
```
JonathanHUnity: What does it mean that only ground truth images are supported?

notebooks/Perception_Statistics.ipynb
```diff
@@ -294,13 +295,17 @@
...
     "\n",
     "bounding_box_definition_id = \"f9f22e05-443f-4602-a422-ebe4ea9b55cb\"\n",
-    "line_width = 5\n",
     "def draw_bounding_boxes(index):\n",
     "    filename = os.path.join(data_root, box_captures.loc[index, \"filename\"])\n",
     "    annotations = box_captures.loc[index, \"annotation.values\"]\n",
+    "\n",
+    "    label_mapping = dict()\n",
+    "    for i in ann_def.get_definition(bounding_box_definition_id)['spec']:\n",
+    "        label_mapping[i['label_id']] = i['label_name']\n",
```
JonathanHUnity: Hmm, I guess this broke when the label names were removed from the bboxes. Seems ok for now, but it would be nice to find a cleaner way to do this mapping.

JonathanHUnity (commented):
Overall looks quite good. I guess we should have someone like @bowenlee919  take a look before merging.

datasetinsights/stats/visualization/bbox3d_plot.py
StevenBorkman: It's the formula for perspective divide which is used to map cartesian coordinates to homogeneous/projected coordinates, see: https://www.scratchapixel.com/lessons/3d-basic-rendering/computing-pixel-coordinates-of-3d-point/mathematics-computing-2d-coordinates-of-3d-points.

The theory is, from my reading, that divide by 0 should always be prevented by the near clipping plane, but it is a good catch and we should protect against that case.

datasetinsights/datasets/synthetic.py
StevenBorkman: Added.

datasetinsights/stats/visualization/plots.py
StevenBorkman: So the other plot bboxes methods have documentation about how the method performs based on if you are using it for 'ground truth' vs. 'prediction.' I'm not sure what would have to change to support prediction, so I wanted to document that. Reworded comment to hopefully clear that up.

datasetinsights/stats/visualization/bbox3d_plot.py
JonathanHUnity: Got it! I learned something today.

JonathanHUnity (dismissed):

datasetinsights/datasets/synthetic.py
```diff
@@ -8,14 +8,15 @@
...
+from pyquaternion import Quaternion
 from sklearn.model_selection import train_test_split
 
 from datasetinsights.datasets.unity_perception import (
     AnnotationDefinitions,
     Captures,
 )
 from datasetinsights.datasets.unity_perception.tables import SCHEMA_VERSION
-from datasetinsights.io.bbox import BBox2D
+from datasetinsights.io.bbox import BBox2D, BBox3d
```
adason: [nit] Since we already have `BBox2D` class, maybe it make sense to change 3D bounding box class to `BBox3D`?

datasetinsights/datasets/synthetic.py
adason: @StevenBorkman Not directly related to this review, but I think we should update this schema [doc](https://github.com/Unity-Technologies/com.unity.perception/blob/master/com.unity.perception/Documentation~/Schema/Synthetic_Dataset_Schema.md#2d-bounding-box) which describes perception SDK's 3D bounding box format.

datasetinsights/io/bbox.py
```diff
@@ -219,7 +219,7 @@ def back_left_bottom_pt(self):
         Returns: :py:class:`float`: Back-left-bottom point.
 
         """
-        p = np.array([-self.length / 2, -self.width / 2, -self.height / 2])
+        p = np.array([-self.width / 2, -self.height / 2, -self.length / 2])
```
adason, line 222: Should have unittests to make sure all corners are calculated correctly.

datasetinsights/stats/visualization/bbox2d_plot.py
```diff
@@ -5,6 +5,7 @@
 """
 import os as _os
 import pathlib
+import random as _random
```
adason: [nit] This is not a common python pattern. You should simply use

```
import random
import cv2
import numpy as np  # use abbreviated alias np is very common 

random.choice(...)
cv2.XXX(...)
np.XXX(...)
```

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -0,0 +1,160 @@
+""" Helper bounding box 3d library to plot pretty 3D boundign
+boxes with a simple Python API.
+"""
+
+import cv2 as _cv2
```
adason: use `cv2` and `np` directly.

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -0,0 +1,160 @@
...
+        back_upper_right (int tuple): Back bottom left coordinate of the 3D
+        bounding box
+        back_bottom_right (int tuple): Back bottom left coordinate of the 3D
+        bounding box
+        color (tuple): RGBA color of the bounding box. Defaults to None. If
+        color = None the the tuple of [0, 255, 0, 255] (Green) will be used.
+        box_line_width: The width of the drawn box. Defaults to 2.
+    """
+    try:
+        fbl = (int(front_bottom_left[0]), int(front_bottom_left[1]))
```
adason: Why is 3D bounding box encoded as integer? In 2D bounding boxes, they are integer because it's in 2D image pixels coordinate. 3D bounding box should support any format in float? 

[update]
I think the confusion originated from the projection from 3D space of 8 corners to 2D pixel space. This should be well documented in the docstring of this method to indicate that they are in image pixel space. Also, I don't think the float to int conversion should be  here as it was already handled by `_project_pt_to_pixel_location`.

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -0,0 +1,160 @@
...
+        bll_raster,
+        bul_raster,
+        bur_raster,
+        blr_raster,
+        color,
+        box_line_width,
+    )
+
+
+def _project_pt_to_pixel_location(pt, projection, img_height, img_width):
```
adason, line 131: Should have unittest for this method.

datasetinsights/stats/visualization/plots.py
```diff
@@ -95,7 +98,9 @@ def _process_label(bbox, label_mappings=None):
     """
     if label_mappings is not None:
         label = label_mappings[bbox.label]
-    if bbox.score != 1.0:
+    else:
+        label = None
```
adason: This should be `label=bbox.label` and you don't need to handle the case when `label=None` in line 103

datasetinsights/stats/visualization/plots.py
```diff
@@ -128,6 +133,39 @@ def get_color(match):
...
+        colors (list): a color list for boxes. Defaults to none. If
+        colors = None, it will default to coloring all boxes green.
+
+    Returns:
+        PIL image: a PIL image with bounding boxes drawn on it.
+    """
+    np_image = np.array(image)
+    img_height, img_width, _ = np_image.shape
+
+    if projection is None:
```
adason: I think `plot_bboxes3d` should fail if `projection` is not provided. The behavior of "plot" is undefined as there are no way to project corners in 3D space to 2D image pixels. Let's make `projection` a required argument here.

adason (changes requested):

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -0,0 +1,160 @@
...
+    """ Projects a 3D coordinate into a pixel location.
+
+    Applies the passed in projection matrix to project a point from the camera's
```
adason, line 134: I am a little rusty on the math that project a point from 3D to 2D image space. Can you add a wiki links or some textbook reference that explain this projection?

datasetinsights/datasets/synthetic.py
StevenBorkman: Ok, it was a class that was already there, so I originally left it as is. But I have updated.

datasetinsights/stats/visualization/bbox3d_plot.py
StevenBorkman: updated

datasetinsights/stats/visualization/bbox3d_plot.py
StevenBorkman: updated.

datasetinsights/stats/visualization/bbox3d_plot.py
StevenBorkman: added

datasetinsights/io/bbox.py
StevenBorkman: added

adason (approved):
Look good. Thanks for helping out with 3D bounding box annotations! 💯

PR #110, opened 2020-10-09
Github Actions Update

Saurav-D:
# Peer Review Information

The `set-env` and `add-path` command is deprecated and will be disabled soon. Link: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/


adason (approved):
Thanks for fixing this 💯

Issue #114, opened 2020-10-20
FileExistsError when writing metrics


Saurav-D:
**Describe the Bug:**
I get file exists error when I try to save metrics on local file system on Kubeflow. This error hangs the kubeflow training pipeline.

**How to Reproduce?**
Try to run train pipeline with `tb_log_dir` as `/run/tb/`

**What did you expect to happen:**
Complete training.

**Console logs / stack traces**
```
KeyError: '/runs/465893c5-8006-4aa6-bf7d-125e4d58d5d5/tb/val/APIOU50-per-class/drink_whippingcream_lucerne'
During handling of the above exception, another exception occurred:
fw = FileWriter(logdir=fw_tag)
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/writer.py", line 95, in __init__
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    directory_check(self._logdir)
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/record_writer.py", line 51, in directory_check
    logdir, max_queue, flush_secs, filename_suffix)
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/event_file_writer.py", line 104, in __init__
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    os.makedirs(path)
  File "/usr/lib/python3.7/os.py", line 221, in makedirs
    exec(code, run_globals)    
directory_check(self._logdir)  File "datasetinsights/__main__.py", line 36, in <module>
  File "/usr/local/lib/python3.7/dist-packages/tensorboardX/record_writer.py", line 51, in directory_check
    entrypoint()
  File "/usr/local/lib/python3.7/dist-packages/click/core.py", line 829, in __call__
    os.makedirs(path)
  File "/usr/lib/python3.7/os.py", line 221, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/runs/465893c5-8006-4aa6-bf7d-125e4d58d5d5/tb/val/APIOU50-per-class/drink_orangejuice_minutemaid'
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/runs/465893c5-8006-4aa6-bf7d-125e4d58d5d5/tb/val/APIOU50-per-class/drink_orangejuice_minutemaid'
```

Saurav-D:
[...]

PR #115, opened 2020-10-26
Remove tensorboardx and Updating KFP

Saurav-D:
# Peer Review Information

Removing tensorboardx library and using `torch.utils.tensorboard`. Updated kubeflow pipeline for starting tensorboard in kubeflow dashboard.

Solves https://github.com/Unity-Technologies/datasetinsights/issues/114 and https://github.com/Unity-Technologies/datasetinsights/issues/109

# Pull Request Check List


datasetinsights/estimators/faster_rcnn.py
```diff
@@ -79,7 +79,7 @@ def __init__(
         self._init_distributed_mode()
         self.no_cuda = no_cuda
         self._init_device()
-        self.writer = SummaryWriter(writer.logdir, write_to_disk=is_master())
+        self.writer = SummaryWriter(logdir)
```
BlairLee, line 82: Does this `SummaryWriter` solve the `connection error` issue?

datasetinsights/io/kfp_output.py
```diff
@@ -68,3 +76,15 @@ def write_metric(self):
         logger.debug(
             f"Metrics file {self.filename} saved at path:" f" {self.filepath}"
         )
+
+    def create_tb_visualization_json(self):
```
BlairLee, line 80: Does it dump the metrics into a json file? Why don't we need it before?

datasetinsights/estimators/faster_rcnn.py
Saurav-D: Yes, it does.

datasetinsights/io/kfp_output.py
Saurav-D: Yes. Previously, we didn't have tensorboard visualization with in kubeflow dashboard which is a pretty useful addition to have.

BlairLee (approved):

datasetinsights/estimators/faster_rcnn.py
adason: @Saurav-D This seems to be enable tensorboard logs for non-master training processes. What's the mechanism behind the scene that prevents non-master training processes from writing to tensorboard?

adason: If the answer is "yes, pytorch tensorboard will handle this behind the scene", this should be moved to `create_estimator` in base.py

Saurav-D: Sure, I can create a PR for this.

datasetinsights/estimators/base.py
```diff
@@ -35,17 +33,18 @@ def create_estimator(
...
     #  writer of non-master processes which could make debugging harder
 
-    writer = SummaryWriter(tb_log_dir)
     kfp_writer = KubeflowPipelineWriter(
-        filename=kfp_metrics_filename, filepath=kfp_metrics_dir,
+        tb_log_dir=tb_log_dir,
+        filename=kfp_metrics_filename,
+        filepath=kfp_metrics_dir,
     )
+    kfp_writer.create_tb_visualization_json()
```
adason, line 41: @Saurav-D Why do we need to call this when the `KubeflowPipelineWriter` is initialized? Can you hide this so that users does not have to remember calling this? How is this different from `KubeflowPipelineWriter.write_metric()`? Should they be coupled into the same writer?

datasetinsights/io/kfp_output.py
```diff
@@ -68,3 +76,15 @@ def write_metric(self):
...
+        try:
+            metadata = {
+                "outputs": [{"type": "tensorboard", "source": self.tb_log_dir}]
+            }
+            with open("/mlpipeline-ui-metadata.json", "w") as f:
```
adason, line 85: @Saurav-D Why is this path hard coded? Can user specify this in CLI so that they can parameterize this in Kubeflow pipeline?

datasetinsights/io/kfp_output.py
```diff
@@ -68,3 +76,15 @@ def write_metric(self):
...
+                json.dump(metadata, f)
+
+        # when we don't have write permission
+        except IOError:
```
adason, line 89: If the output json path is not hard coded, this check can be removed. It's not very good practice if this failed silently.

datasetinsights/io/kfp_output.py
Saurav-D: This needs to be hardcoded because kubeflow wants it to be `/mlpipeline-ui-metadata.json` it does not works if the path is changed.

datasetinsights/io/kfp_output.py
Saurav-D: The output path has to be hardcoded.

datasetinsights/estimators/base.py
Saurav-D: `write_metrci()` creates JSON file for metrcis while this creates JSON for tensorboard visualizer. I think they should be coupled to the same writer because `KubeflowPipelineWriter ` shuold handle anything that is related to KFP pipeline/ dashboard.

Saurav-D: `kfp_writer.create_tb_visualization_json()` we can hide this and instead call when the writer is initialized so that the user don't have to worry about it.

datasetinsights/io/kfp_output.py
Saurav-D: Also, the reason for this was when we run this locally and we don't have write permission on `/` which will cause it to fail which we ideally don't want as writing this file is not necessary when running locally instead of kubeflow.

Issue #112, opened 2020-10-16
Can't build Docker Image Locally


Saurav-D:
**Describe the Bug:**
When trying to build docker image locally, even without any changes, build fails.

**How to Reproduce?**
Get master branch of the repo and try to run `docker build`

**What did you expect to happen:**
Build Successfully.

**Console logs / stack traces**
```
Step 12/14 : RUN poetry config virtualenvs.create false     && poetry install --no-dev
 ---> Running in 8ea17e71488d
Traceback (most recent call last):
  File "/usr/local/bin/poetry", line 7, in <module>
    from poetry.console import main
  File "/usr/local/lib/python3.7/dist-packages/poetry/console/__init__.py", line 1, in <module>
    from .application import Application
  File "/usr/local/lib/python3.7/dist-packages/poetry/console/application.py", line 7, in <module>
    from .commands.about import AboutCommand
  File "/usr/local/lib/python3.7/dist-packages/poetry/console/commands/__init__.py", line 4, in <module>
    from .check import CheckCommand
  File "/usr/local/lib/python3.7/dist-packages/poetry/console/commands/check.py", line 2, in <module>
    from poetry.factory import Factory
  File "/usr/local/lib/python3.7/dist-packages/poetry/factory.py", line 18, in <module>
    from .repositories.pypi_repository import PyPiRepository
  File "/usr/local/lib/python3.7/dist-packages/poetry/repositories/pypi_repository.py", line 33, in <module>
    from ..inspection.info import PackageInfo
  File "/usr/local/lib/python3.7/dist-packages/poetry/inspection/info.py", line 25, in <module>
    from poetry.utils.env import EnvCommandError
  File "/usr/local/lib/python3.7/dist-packages/poetry/utils/env.py", line 22, in <module>
    import virtualenv
  File "/usr/local/lib/python3.7/dist-packages/virtualenv/__init__.py", line 3, in <module>
    from .run import cli_run, session_via_cli
  File "/usr/local/lib/python3.7/dist-packages/virtualenv/run/__init__.py", line 5, in <module>
    from ..app_data import AppDataAction, AppDataDisabled, TempAppData
  File "/usr/local/lib/python3.7/dist-packages/virtualenv/app_data/__init__.py", line 10, in <module>
    from appdirs import user_data_dir
ModuleNotFoundError: No module named 'appdirs'
The command '/bin/sh -c poetry config virtualenvs.create false     && poetry install --no-dev' returned a non-zero code: 1

[...]

PR #116, opened 2020-10-27
Docker Issue Fix

Saurav-D:
# Peer Review Information

Solves https://github.com/Unity-Technologies/datasetinsights/issues/112

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,3 @@
-# FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
 FROM nvidia/cuda:11.1-cudnn8-devel-ubuntu18.04
 
 RUN apt-get update \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
+# FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
+FROM nvidia/cuda:11.1-cudnn8-devel-ubuntu18.04
 
 RUN apt-get update \
     && apt-get install -y \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
+# FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
```
BlairLee: Can we remove this comment?

BlairLee:
Great job!

Dockerfile
Saurav-D: Sure, sorry forgot that.

BlairLee (approved):

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM nvidia/cuda:10.0-cudnn7-runtime-ubuntu18.04
+FROM nvidia/cuda:11.1-cudnn8-devel-ubuntu18.04
```
adason, line 1: @Saurav-D Please revert this. This change will disable GPU support for pytorch as the package in pypi only supports CUDA 10 for pytorch 1.4.x.

pyproject.toml
```diff
@@ -99,5 +99,5 @@ testpaths = [
 datasetinsights = "datasetinsights.__main__:entrypoint"
 
 [build-system]
-requires = ["poetry>=1.0.5"]
+requires = ["poetry>=1.0.10"]
```
adason, line 102: @Saurav-D This does not pin poetry version to 1.0.10.

Dockerfile
Saurav-D: @adason Is there any issue about this on pytorch? I launched train pipeline using this and it ran successfully using GPUs.

pyproject.toml
Saurav-D: Yes. It's working without pinning poetry version.

Dockerfile
Saurav-D: Also, pip install downgrades cuda version to 10.1 as can be seen in the following:
```
# python
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torch
>>> print(torch.version.cuda)
10.1
```

I ran the above inside the docker image of the latest build.

PR #122, opened 2020-11-06
Docker Github Actions Update

Saurav-D:
# Peer Review Information

Docker Build Push github  actions repo is updated to v2. https://github.com/Unity-Technologies/datasetinsights/actions/runs/348350305

### DIFF
--- a/.github/workflows/publish-docker-hub.yaml
+++ b/.github/workflows/publish-docker-hub.yaml
@@ -10,7 +10,7 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2
       - name: Push to Docker Hub
-        uses: docker/build-push-action@v1
+        uses: docker/build-push-action@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-docker-hub.yaml
```diff
@@ -10,7 +10,7 @@ jobs:
       - name: Check out the repo
         uses: actions/checkout@v2
       - name: Push to Docker Hub
-        uses: docker/build-push-action@v1
```
adason, line 13: This is less critical change. I was wondering what's changed in v2.

adason:
It seems v1-v2 upgrade require backward non-compatible changes. I think we should keep v1 until we are ready to move to v2 syntax.  

https://github.com/docker/build-push-action/blob/master/UPGRADE.md

Close this PR.

PR #123, opened 2020-11-10
Mlflow tracker

86sanj:
# Peer Review Information

MLflow Tracking logging code. This includes following changes:
1. TrackerFactory: Creates and returns tracker singleton instance based on identifier. 
2. MLFlowTracker: It initializes mlflow with host&client id and starts background thread. 
3. NullTracker: To disable tracking
4. FasterRCNN: MLflow Tracking logs.
5. test_faster_rcnn: Fixes of existing UT and added few test check.
6. test_mlflow_tracker: Holds unit test case of TrackerFactory, MLFlowTracker, NullTracker.

# Pull Request Check List


datasetinsights/constants.py
```diff
@@ -53,3 +53,6 @@
     "allow_extra_args": True,
 }
 DEFAULT_DATASET_VERSION = "latest"
+
+MLFLOW_TRACKER = "mlflow"
+DEFAULT_RUN_NAME = "run_" + str(datetime.now())
```
Saurav-D: Can this name be more specific as to which run it is referring to?

datasetinsights/configs/faster_rcnn_fine_tune.yaml
```diff
@@ -42,3 +42,8 @@ optimizer:
 pretrained: False
 pretrained_backbone: True
 synchronize_metrics: True
+mlflow:
```
adason: If I didn't have mlflow section in my config yaml, what will happen? This seems to require all future yaml to include mlflow section which is not desirable. 

Also, is this config required for every estimator yaml? Do they need to be repeated for every estimator yaml? Can they be shared across estimators?

datasetinsights/configs/faster_rcnn_groceries_real.yaml
```diff
@@ -3,7 +3,7 @@ backbone: resnet50
 num_classes: 64
 task: object_detection
 train:
-  epochs: 100
+  epochs: 10
```
adason: Why is this change to 10 instead of the default 100?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
+import os
```
adason: Why is this file called mzflow? Isn't mlflow a better name? Is this MLFlow specific convention?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+
+import mlflow
+from google.auth.transport.requests import Request
+from google.oauth2 import id_token
+
+MLFLOW = "mlflow"
```
adason: These should be class constant variables.

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,19 @@
+from datasetinsights.constants import MLFLOW_TRACKER
+from datasetinsights.io.tracker.mzflow import MLFlowTracker
+
+
+class TrackerFactory:
```
adason, line 10: Why do we need a factory class with just a method create? Can this be a simple python function such as `create_tracker`?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+HOST_ID = "tracking_host_id"
+NAME = "experiment_name"
+DEFAULT_NAME = "datasetinsights"
+
+
+class MLFlowTracker:
+    """MlFlow tracker class, responsible for creating singleton object,
+    refresh token and get instance .
+
+    Attributes:
```
adason: Probably don't need to document private variable. In general you are hiding this from users.

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+    __instance = None
+    __client_id = None
+
+    def __init__(self, client_id):
+        """Virtually private constructor.
+
+        Args:
+            client_id : MLFlow tracking server client id
+        """
+        if MLFlowTracker.__instance:
```
adason: It's desirable if someone calls the constructor of a singleton object, it will directly return the instance instead of raising exceptions.

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+            raise Exception("This class is a singleton!")
+        else:
+            if client_id:
+                self.__client_id = client_id
+            MLFlowTracker.__instance = self
+
+    @staticmethod
+    def get_instance(client_id=None):
+
```
adason: Why empty line here?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+MLFLOW = "mlflow"
+CLIENT_ID = "tracking_client_id"
+HOST_ID = "tracking_host_id"
+NAME = "experiment_name"
+DEFAULT_NAME = "datasetinsights"
+
+
+class MLFlowTracker:
+    """MlFlow tracker class, responsible for creating singleton object,
```
adason: Might be helpful to write sample use cases code snippet in docstring.

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+        Args:
+            client_id : MLFlow tracking server client id
+        """
+
+        if not MLFlowTracker.__instance:
+            MLFlowTracker(client_id)
+        return MLFlowTracker.__instance
+
+    @staticmethod
+    def get_tracker(config):
```
adason: I am a little confused by `get_instance` and `get_tracker`. Also, why is this method not called in the __init__ method?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+
+        """get instance of tracker class.
+
+        Args:
+            config : config object, holds server details
+        """
+        client_id = config.get(MLFLOW).get(CLIENT_ID, None)
+        host_id = config.get(MLFLOW).get(HOST_ID, None)
+        exp_name = config.get(MLFLOW).get(NAME, DEFAULT_NAME)
+        if host_id:
```
adason: The nested if here seems can be flattened. Also, what will happen in the following cases:
1. either host_id or client_id is missing
2. both host_id and client_id is missing

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+            if client_id:
+                MLFlowTracker.refresh_token(client_id)
+            mlflow.set_tracking_uri(host_id)
+            mlflow.set_experiment(experiment_name=exp_name)
+        config.pop(MLFLOW, None)
```
adason: It's very dangerous to change the config yaml as a side effect of this function, without copying the config and do some operations. Can this side effect be avoided?

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+        return mlflow
+
+    @staticmethod
+    def refresh_token(client_id=None):
```
adason: Does this method tries to fetch token based on GCP credential? What's the desired behavior if client_id is not provided?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -110,6 +114,7 @@ def __init__(
 
         if checkpoint_file:
             self.checkpointer.load(self, checkpoint_file)
+        self.mltracking = TrackerFactory.create(config, const.MLFLOW_TRACKER)
```
adason: I will suggest using explicit name such as `mlflow_tracking`. It's not clear to readers when `mltracking` is used, as kubeflow and tensorboard are both mltracking stuff.

datasetinsights/constants.py
```diff
@@ -53,3 +53,6 @@
     "allow_extra_args": True,
 }
 DEFAULT_DATASET_VERSION = "latest"
+
+MLFLOW_TRACKER = "mlflow"
```
adason: These variables are better stored in `io/tracker/factory.py` as you will have `KUBEFLOW_TRACKER`, `TENSORBOARD_TRACKER` in the future.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -20,6 +20,8 @@
 from datasetinsights.evaluation_metrics.base import EvaluationMetric
 from datasetinsights.io.bbox import BBox2D
 from datasetinsights.io.summarywriter import get_summary_writer
+from datasetinsights.io.tracker.factory import TrackerFactory
+from datasetinsights.io.tracker.mzflow import MLFlowTracker
```
adason: Why is this import required if the TrackerFactory can provide an instance of MLflow tracking object?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -419,12 +435,15 @@ def evaluate_per_epoch(
             loss_dict_reduced = reduce_dict(loss_dict)
             losses = sum(loss for loss in loss_dict_reduced.values())
             loss_metric.update(avg_loss=losses.item(), batch_size=len(targets))
+            # refresh mlflow token
+            MLFlowTracker.refresh_token()
```
adason: It's not clear why this has to be called in faster_rcnn.py. Can this be handled by `MLFlowTracker` class itself?

adason (changes requested):

datasetinsights/estimators/faster_rcnn.py
86sanj: Token expires after 4 or 5 epochs, and throws expedition:
mlflow.exceptions.MlflowException: API request to endpoint /api/2.0/mlflow/runs/log-metric failed with error code 401 != 200. Response body: 'OpenID Connect token expired: JWT has expired'

-  So we need to refresh the token, evaluate_per_epoch is common in train and evaluate so refreshing token at this place.

datasetinsights/estimators/faster_rcnn.py
86sanj: MLFlowTracker : This import requires to refresh the token.
TrackerFactory: Requires to get tracker instance.

datasetinsights/io/tracker/mzflow.py
86sanj: mlflow is the official module name, so for better readability changed the name. Its not a convention.

datasetinsights/configs/faster_rcnn_groceries_real.yaml
86sanj: Fixed!

datasetinsights/io/tracker/mzflow.py
86sanj: Fixed!

datasetinsights/io/tracker/mzflow.py
86sanj: Fixed!

datasetinsights/io/tracker/mzflow.py
86sanj: Fixed!

datasetinsights/constants.py
86sanj: Fixed

datasetinsights/configs/faster_rcnn_fine_tune.yaml
86sanj: If I didn't have mlflow section in my config yaml, what will happen? This seems to require all future yaml to include mlflow section which is not desirable.
- In current implementation this is required, but I can add some validation to handle this. 
When we add enable disable feature this can be optional. 

Also, is this config required for every estimator yaml? Do they need to be repeated for every estimator yaml? Can they be shared across estimators?
- If an estimator wants to use MLFlow, it need MLFlow details. As of now we have decided to put MLFlow infra details in config so it need to be in all configs if its used by estimator. 

- It can be shared. I can create separate YAML file which is specific to MLFlow infra. All configs need to import this if they want to use MLFlow.

datasetinsights/estimators/faster_rcnn.py
adason: Can the token refresh handled by MLFlowTracker class itself? Why do user need to remember refreshing tokens for MLFlow tracker?

datasetinsights/estimators/faster_rcnn.py
adason: Related to `refresh_token` discussion below.

datasetinsights/io/tracker/factory.py
86sanj: We need factory method to create multiple trackers such as MLFlowTracker, KubeflowTracker, TensorboardTracker.

datasetinsights/configs/faster_rcnn_fine_tune.yaml
adason: Since the yaml file with mlflow section will not be required in the future, they should be disabled by default. I still think it's valuable to implement `NullTracker` so that it's used by default when user does not specify mlflow in their yaml file.

romerocesar: also, for consistency, we should probably rename this as the `tracker` section and there can be settings inside for mlflow, tensorboard, comet, wandb in the future

datasetinsights/io/tracker/mzflow.py
romerocesar: how does this improve readability and what's the significance of the z? it's clearly a different import path so I don't think having the same module name is a big problem.

datasetinsights/io/tracker/factory.py
romerocesar: The question is specific to why a class, not why a factory.

I think the answer may be for singletons. python supports closures which makes classes not strictly required for this, but it's probably more readable than wondering how the function is able to get a hold of a previously instantiated tracker

datasetinsights/estimators/faster_rcnn.py
romerocesar: also, for consistency, use tracker instead of tracking here

datasetinsights/estimators/faster_rcnn.py
romerocesar: as we discussed, let's leave the responsibility of refreshing the token entirely to the tracker object and not the client (i.e. estimator in this case)

datasetinsights/configs/faster_rcnn_fine_tune.yaml
```diff
@@ -42,3 +42,8 @@ optimizer:
...
+  experiment_name: datasetinsights
+  run_name:
+  tracking_client_id:
+  tracking_host_id:
```
romerocesar: since these are inside the mlflow block, we can simplify the names to improve readability. For example:
```yaml
experiment:
run:
client_id:
host:
```

datasetinsights/constants.py
romerocesar: I think we should take this out of this global constants file. This is fairly specific to mlflow anyways, so put this in the tracker factory or the mlflow tracker itself.

Also, a better strategy to generate ids that benefit from being human readable is something like `202011121633-foo-bar` where the first token is just a readable datetime and the second a random adjective and random object which makes the id much easier to share, find and remember.

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -192,12 +197,14 @@ def train(self, train_data, val_data=None, **kwargs):
         val_loader = dataloader_creator(
             config, val_dataset, val_sampler, VAL, self.distributed
         )
-        self.train_loop(
-            train_dataloader=train_loader,
-            label_mappings=label_mappings,
-            val_dataloader=val_loader,
-            train_sampler=train_sampler,
-        )
+        with self.mltracking.start_run(run_name=const.DEFAULT_RUN_NAME):
```
romerocesar: - why doesn't the run start when the mlflow tracker instance is created?
- why does this ignore the run name that may be passed by the user in a config or CLI flag? this part in particular should include a unit test

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -442,6 +461,7 @@ def log_metric_val(self, label_mappings, epoch):
             if metric.TYPE == "scalar":
                 self.writer.add_scalar(f"val/{metric_name}", result, epoch)
                 self.kfp_writer.add_metric(name=metric_name, val=result)
+                self.mltracking.log_metric(f"val/{metric_name}", result)
```
romerocesar: yikes we definitely need that tracker interface sooner than later

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,19 @@
...
+    """Factory: responsible for creating instance of tracker class"""
+
+    @staticmethod
+    def create(config, tracker_type):
```
romerocesar: use named params instead of positional ones

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,19 @@
...
+        """create tracker class object.
+
+        Args:
+            config : config object, holds server details
+            tracker_type : type of tracker
+        """
+        if MLFLOW_TRACKER == tracker_type:
+            return MLFlowTracker.get_tracker(config)
```
romerocesar: remember to change this to call the CTOR with the client_id etc. so the factory keep the reference to the singleton rather than the tracker itself doing that

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+    refresh token and get instance .
+
+    Attributes:
+        __instance: holds singleton instance
+        __client_id: holds ml tracking server client id
+
+    """
+
+    __instance = None
+    __client_id = None
```
romerocesar: these will probably become instance attributes initialized in the CTOR (i.e. `self.client_id`) now that the tracker will hold the thread that refreshes its credentials behind the scenes

datasetinsights/io/tracker/mzflow.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def __init__(self, client_id):
+        """Virtually private constructor.
```
romerocesar: what do you mean by this? there is no such thing in python AFAIK

datasetinsights/io/tracker/mzflow.py
romerocesar: just move the singleton logic to the factory

tests/test_faster_rcnn.py
```diff
@@ -46,8 +46,10 @@ def config():
     return cfg
 
 
-def test_faster_rcnn_train_one_epoch(config, dataset):
+@patch("datasetinsights.io.tracker.factory.TrackerFactory.create")
+def test_faster_rcnn_train_one_epoch(mock_create, config, dataset):
```
romerocesar, line 51: why should this test be changed rather than a different test be created that actually checks the factory is called to get a tracker?

tests/test_faster_rcnn.py
```diff
@@ -95,10 +97,12 @@ def test_faster_rcnn_train_one_epoch(config, dataset):
 
 @patch("datasetinsights.estimators.faster_rcnn.FasterRCNN.train_one_epoch")
 @patch("datasetinsights.estimators.faster_rcnn.Loss.compute")
+@patch("datasetinsights.io.tracker.factory.TrackerFactory.create")
 def test_faster_rcnn_train_all(
-    mock_loss, mock_train_one_epoch, config, dataset
+    mock_create, mock_loss, mock_train_one_epoch, config, dataset
 ):
     """test train on all epochs."""
+    mock_create.return_value = MagicMock()
```
romerocesar, line 106: same as above

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+@pytest.fixture
+def config():
+    """prepare config."""
+    with open("tests/configs/faster_rcnn_groceries_real_test.yaml") as f:
+        cfg = CN.load_cfg(f)
+
+    return cfg
+
+
+def test_get_instance():
```
romerocesar: update this to test the singleton in the factory

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+    first_instance = MLFlowTracker.get_instance()
+    second_instance = MLFlowTracker.get_instance()
+    assert first_instance == second_instance
+
+
+@patch("datasetinsights.io.tracker.mzflow.mlflow")
+@patch("datasetinsights.io.tracker.mzflow.MLFlowTracker.refresh_token")
```
romerocesar: please avoid the confusing name mzflow

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+    MLFlowTracker.get_tracker(config)
+    mock_refresh.assert_called_with(CLIENT_ID)
+    mock_mlflow.set_tracking_uri.assert_called_with(HOST_ID)
+    mock_mlflow.set_experiment.assert_called_with(experiment_name=EXP_NAME)
+    assert not config.get("mlflow")
+
+
+@patch("datasetinsights.io.tracker.mzflow.id_token.fetch_id_token")
+@patch("datasetinsights.io.tracker.mzflow.MLFlowTracker.get_instance")
+def test_refresh_token(mock_get_instance, mock_id_token):
```
romerocesar: remember to remove this since a thread will take care of refreshing the token in the bg

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,59 @@
...
+    mock_get_instance.return_value = tracker
+    mock_id_token.return_value = TEST_TOKEN
+    MLFlowTracker.refresh_token(CLIENT_ID)
+    mock_get_instance.assert_called_with(CLIENT_ID)
+
+
+@patch("datasetinsights.io.tracker.mzflow.MLFlowTracker.get_tracker")
+def test_create_tracker_factory(mock_get_tracker, config):
+    TrackerFactory.create(config, MLFLOW_TRACKER)
+    mock_get_tracker.assert_called_with(config)
```
romerocesar: rename test to not suggest it's creating a factory

romerocesar (changes requested):

datasetinsights/io/tracker/mzflow.py
86sanj: Yes, service account need to set in environment variable. If clientId is not there then you will get following error
- json.decoder.JSONDecodeError: Expecting value: line 2 column 1 (char 1) when you initialize mlflow :mlflow.set_experiment(experiment_name=exp_name)

Ref: https://googleapis.dev/python/google-auth/1.14.0/_modules/google/oauth2/id_token.html#fetch_id_token

tests/test_faster_rcnn.py
86sanj: I am only making change to mock factory.create call because of following two reasons: 
1. We are creating estimator object under this method which calls factory.create so do want to make call to outside of the estimator constructor.
2. Also, under train_one_epoch, I am doing mltracking logging and I don't want to do logging when I run unit test case so need to mock this call.

datasetinsights/estimators/faster_rcnn.py
86sanj: mlflow run starts when your experiment run starts and it ends when your experiment run end.
 
There are two ways to start mlflow run.
1. Using "with" block: mlflow ends the runs automatically when you exit the with block.
2. using manual start and manual end
Documentation: https://www.mlflow.org/docs/latest/python_api/mlflow.html#mlflow.start_run

- I can start the run when tracker instance gets created, this will allow to change run name from CLI. But I need to put train/eval run under try block and end mlflow run in catch/finally block and change the status of run according to fail or success of experiment run.

datasetinsights/io/tracker/mzflow.py
86sanj: There are 5 possibilities:

Case1: When there is no tracker tag in YAML: 
Returns NullMLFlowTracker or mlflow which logs in local file and experiment run as usual 

Case2: When there is tracker tag but there is no mlflow tag in YAML: 
Returns NullMLFlowTracker or mlflow which logs in local file and experiment run as usual 

Case3: When there are tracker, mlflow tag but host is not configured in YAML: 
Returns NullMLFlowTracker or mlflow which logs in local file and experiment run as usual 

Case4: When there are tracker, mlflow, host tag but not client is not configured in YAML: 
Returns return Mlflow with given host initialization, end user may or may not have client id setup

Case5: When there are tracker, Mlflow, host , client ID configured in YAML: 
Returns return Mlflow with given host client id, initialization, background thread(daemon thread ) running in background which reset token at every 8 minutes

datasetinsights/constants.py
86sanj: Changed with this kind of name: train_20201113-162123, eval_20201113-162125

datasetinsights/io/tracker/mzflow.py
romerocesar: what's the behavior in the estimator though? it should catch this exception and log a warning since the job can still run just without tracking

datasetinsights/constants.py
romerocesar: why not use dash everywhere? instead of mixing `-` and `_`?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -31,6 +32,7 @@
 TRAIN = "train"
 VAL = "val"
 TEST = "test"
+NAME = "run_name"
```
romerocesar: add a comment that explains what these generic constants are for, and probably should add `__` prefix since they're not meant to be used outside this module

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -192,12 +197,23 @@ def train(self, train_data, val_data=None, **kwargs):
...
-        self.train_loop(
-            train_dataloader=train_loader,
-            label_mappings=label_mappings,
-            val_dataloader=val_loader,
-            train_sampler=train_sampler,
-        )
+
+        params = self.config.clone()
+        if params.get(TrackerFactory.TRACKER):
+            params.pop(TrackerFactory.TRACKER)
```
romerocesar, line 205: *fixme*
what's the point of poping this key from the dictionary? it may result in some function logging a config that is different than the one actually used

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -192,12 +197,23 @@ def train(self, train_data, val_data=None, **kwargs):
...
+            self.train_loop(
+                train_dataloader=train_loader,
+                label_mappings=label_mappings,
+                val_dataloader=val_loader,
+                train_sampler=train_sampler,
+            )
+        except Exception as e:
+
+            self.mlflow_tracker.end_run(status=TrackerFactory.RUN_FAILED)
+            raise e
```
romerocesar, line 218: *fixme*
Log a little something before reraising to at least let us know you're reporting the end of the run as a failure to mlflow

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -314,17 +330,21 @@ def train_one_epoch(
                     intermediate_loss,
                     examples_seen,
                 )
+                self.mlflow_tracker.log_metric(
+                    "training/intermediate_loss", intermediate_loss
```
romerocesar: **fixme**
remove the magic strings from the middle of the code

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -314,17 +330,21 @@ def train_one_epoch(
...
                 optimizer.zero_grad()
-
-        self.writer.add_scalar("training/loss", loss_metric.compute(), epoch)
+        loss_metric_c = loss_metric.compute()
+        self.writer.add_scalar("training/loss", loss_metric_c, epoch)
         self.writer.add_scalar(
             "training/lr", optimizer.param_groups[0]["lr"], epoch
         )
         loss_metric.reset()
+        self.mlflow_tracker.log_metric("training/loss", loss_metric_c)
```
romerocesar: **fixme**: remove magic strings

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -344,12 +364,22 @@ def evaluate(self, test_data, **kwargs):
...
-        self.evaluate_per_epoch(
-            data_loader=test_loader,
-            epoch=0,
-            label_mappings=label_mappings,
-            synchronize_metrics=self.sync_metrics,
-        )
+
+        params = self.config.clone()
+        if params.get(TrackerFactory.TRACKER):
+            params.pop(TrackerFactory.TRACKER)
```
romerocesar, line 372: **fixme:** why pop?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -344,12 +364,22 @@ def evaluate(self, test_data, **kwargs):
...
+        try:
+            self.evaluate_per_epoch(
+                data_loader=test_loader,
+                epoch=0,
+                label_mappings=label_mappings,
+                synchronize_metrics=self.sync_metrics,
+            )
+        except Exception as e:
+            self.mlflow_tracker.end_run(status=TrackerFactory.RUN_FAILED)
+            raise e
```
romerocesar, line 385: **fixme** log more context

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -425,6 +455,7 @@ def evaluate_per_epoch(
         logger.info(f"validation loss is {val_loss}")
 
         self.writer.add_scalar("val/loss", val_loss, epoch)
+        self.mlflow_tracker.log_metric("val/loss", val_loss)
```
romerocesar: wasn't the point of the constants above to use them here?

datasetinsights/io/tracker/factory.py
romerocesar: **ping**

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+                    mlflow_tracker = TrackerFactory._mlflow_tracker_instance(
+                        host_id=host_id, client_id=client_id, exp_name=exp_name
+                    ).get_mlflow()
+                    mlflow_tracker.start_run(
+                        run_name=TrackerFactory.DEFAULT_RUN_NAME
+                    )
+                    return mlflow_tracker
+            return TrackerFactory._null_tracker()
+        else:
+            raise NotImplementedError(f"Unknown tracker {tracker_type}!")
```
romerocesar: I think ValueError is more appropriate. Even better, create an `InvalidTrackerError` so it's very obvious we're raising the exception as opposed to some downstream dependency

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+                    exp_name = mlflow_config.get(TrackerFactory.EXP_NAME, None)
+                    TrackerFactory.update_run_name(mlflow_config)
+                    mlflow_tracker = TrackerFactory._mlflow_tracker_instance(
+                        host_id=host_id, client_id=client_id, exp_name=exp_name
+                    ).get_mlflow()
+                    mlflow_tracker.start_run(
+                        run_name=TrackerFactory.DEFAULT_RUN_NAME
+                    )
+                    return mlflow_tracker
+            return TrackerFactory._null_tracker()
```
romerocesar, line 52: **fixme** each return statement should include a log.info so we can trace what happened when looking at the logs

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+        """private getter method to get singleton instance.
+
+        Returns:
+            NullTracker singleton instance.
+        """
+        if not TrackerFactory.__tracker_instance:
+            with TrackerFactory.__singleton_lock:
+                if not TrackerFactory.__tracker_instance:
+                    TrackerFactory.__tracker_instance = NullTracker()
+        return TrackerFactory.__tracker_instance
```
romerocesar, line 90: can't the null tracker be static since it doesn't need to keep any state?

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+    def update_run_name(mlflow_config):
+
+        """Static method to update run name from config.
+
+        Args:
+            mlflow_config : mlflow_config dictionary
+        """
+        run_name = mlflow_config.get(TrackerFactory.RUN_NAME, None)
+        if run_name:
+            TrackerFactory.DEFAULT_RUN_NAME = run_name
```
romerocesar: **????????????????**

- why does a static method in the factory take in params that look specific to mlflow?
- why override the default name?
- why isn't the name of the run updated through the tracker itself?

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def handle_dummy(self, *args, **kwargs):
+        """method to handle all calls on tracker.
+        """
+        return
+
+    def __getattr__(self, name):
+        """gets called at every call on the instance of this class.
+        """
+        return getattr(self, "handle_dummy")
```
romerocesar: **fixme**
rename to something like stdout handler and just log the input in the handler function before returning so we can at least tell easily when it's being called

datasetinsights/io/tracker/mlflow.py
```diff
@@ -0,0 +1,101 @@
...
+        REFRESH_INTERVAL: default refresh token interval
+        __mlflow: holds initialized mlflow
+    """
+
+    REFRESH_INTERVAL = (
+        3000  # every 50 minutes refresh token. Token expires in 1 hour
+    )
+    __mlflow = None
+
+    def __init__(self, host_id, client_id, exp_name):
```
romerocesar: **fixme** force named params

datasetinsights/io/tracker/mlflow.py
```diff
@@ -0,0 +1,101 @@
...
+        """constructor.
+        Args:
+            host_id: MlTracker server host
+            client_id: MLFlow tracking server client id
+            exp_name: name of the experiment
+        """
+        if client_id:
```
romerocesar, line 61: shouldn't this fail immediately if there is no client id?

datasetinsights/io/tracker/mlflow.py
```diff
@@ -0,0 +1,101 @@
...
+            thread = RefreshTokenThread(client_id)
+            thread.daemon = True
+            thread.start()
+        mlflow.set_tracking_uri(host_id)
+        if exp_name:
+            mlflow.set_experiment(experiment_name=exp_name)
+        self.__mlflow = mlflow
+
+    def get_mlflow(self):
+
```
romerocesar: empty line?

datasetinsights/io/tracker/mlflow.py
```diff
@@ -0,0 +1,101 @@
...
+        """ method to access initialized mlflow
+        Returns:
+            Initialized __mlflow instance.
+        """
+        return self.__mlflow
+
+    @staticmethod
+    def refresh_token(client_id):
```
romerocesar: **fixme**
why is this static and public? shouldn't it keep the token itself instead of or in addition to whatever is happen with the environment variable?

datasetinsights/io/tracker/mlflow.py
```diff
@@ -0,0 +1,101 @@
...
+        >>> mlflow.log_artifact("output.txt", "run1/output/")
+        # ends the run launched under the current experiment
+        >>> mlflow.end_run()
+    Attributes:
+        REFRESH_INTERVAL: default refresh token interval
+        __mlflow: holds initialized mlflow
+    """
+
+    REFRESH_INTERVAL = (
+        3000  # every 50 minutes refresh token. Token expires in 1 hour
```
romerocesar, line 40: did we check whether we can make the token last 1 week?

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,115 @@
...
+
+    return cfg
+
+
+@patch("datasetinsights.io.tracker.mlflow.id_token.fetch_id_token")
+def test_refresh_token(mock_id_token):
+    tracker = MagicMock()
+    tracker.client_id = MagicMock(CLIENT_ID)
+    mock_id_token.return_value = TEST_TOKEN
+    MLFlowTracker.refresh_token(CLIENT_ID)
```
romerocesar: it doesn't make sense for this to be static

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,115 @@
...
+    mock_id_token.assert_called_once()
+
+
+def test_update_run_name(config):
+    config.tracker.mlflow.run = RUN_NAME
+    mlflow_config = config["tracker"].get(TrackerFactory.MLFLOW_TRACKER)
+    TrackerFactory.update_run_name(mlflow_config)
+    assert TrackerFactory.DEFAULT_RUN_NAME == RUN_NAME
```
romerocesar: there are several questions around this `update_run_name` above

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,115 @@
...
+    mock_mlflowtracker.start_run.assert_called_with(run_name=RUN_NAME)
+
+
+@patch("datasetinsights.io.tracker.factory.TrackerFactory._null_tracker")
+def test_factory_create_nulltracker(mock_get_tracker, config):
+    config.tracker.mlflow.client_id = None
+    config.tracker.mlflow.host = None
+    config.tracker.mlflow.experiment = None
+    TrackerFactory.create(config, TrackerFactory.MLFLOW_TRACKER)
+    mock_get_tracker.assert_called_with()
```
romerocesar: called with what? did you mean to say called_once?

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+@patch("datasetinsights.io.tracker.factory.NullTracker.handle_dummy")
+def test__nulltracker(mock_handle_dummy):
```
romerocesar: single `_`

tests/test_mlflow_tracker.py
```diff
@@ -0,0 +1,115 @@
...
+    mock_refresh.assert_called_once()
+    mock_mlflow.set_tracking_uri.assert_called_with(HOST_ID)
+
+
+@patch("datasetinsights.io.tracker.mlflow.RefreshTokenThread.run")
+def test__refresh_token_thread(mock_thread_run):
+    thread = RefreshTokenThread(CLIENT_ID)
+    thread.daemon = True
+    thread.start()
+    mock_thread_run.assert_called_once()
```
romerocesar, line 106: don't you want to test that the thread is actually initialized when the tracker is instantiated?

romerocesar (changes requested):

datasetinsights/configs/faster_rcnn_groceries_real.yaml
```diff
@@ -42,3 +42,9 @@ optimizer:
 pretrained: False
 pretrained_backbone: True
 synchronize_metrics: True
+tracker:
```
adason: The code should not fail without changing the original yaml file. So, why is this change required?

datasetinsights/estimators/faster_rcnn.py
```diff
@@ -192,12 +197,23 @@ def train(self, train_data, val_data=None, **kwargs):
...
+                train_dataloader=train_loader,
+                label_mappings=label_mappings,
+                val_dataloader=val_loader,
+                train_sampler=train_sampler,
+            )
+        except Exception as e:
+
+            self.mlflow_tracker.end_run(status=TrackerFactory.RUN_FAILED)
+            raise e
+        self.mlflow_tracker.end_run()
```
adason: This might be better wrapped into the try/catch/except/finally clause to ensure it was always executed when `self.train_loop(...)` does not raise error.

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+            tracker_type : type of tracker
+
+        Returns:
+            tracker instance.
+        """
+        if TrackerFactory.MLFLOW_TRACKER == tracker_type:
+
+            tracker = config.get(TrackerFactory.TRACKER, None)
+            if tracker and tracker.get(TrackerFactory.MLFLOW_TRACKER, None):
+                mlflow_config = tracker.get(TrackerFactory.MLFLOW_TRACKER)
```
adason, line 36: This is calling for some validation logic against the confit yaml file of the mlflow tracker section. We should provide enough error logging if the config is specified correctly. Otherwise, user will see mysterious error with no clues how to fix the yaml file.

datasetinsights/io/tracker/factory.py
```diff
@@ -0,0 +1,115 @@
...
+        if run_name:
+            TrackerFactory.DEFAULT_RUN_NAME = run_name
+
+
+class NullTracker:
+    """A null tracker that writes nothing. This tracker is
+    used to disable tracking.
+    """
+
+    def handle_dummy(self, *args, **kwargs):
```
adason: [nit] Might be better keeping this method private. It's very unlikely users will need to call this method themself.

adason (changes requested):

datasetinsights/io/tracker/mzflow.py
86sanj: - if client_id is not provided and host is given:
1. If this run is executed by unity AI team then client id is required, in this case  it will throw exception and NullTracker will be initiated.
2. If this run is executed by non unity employ, it will try to connect to given tracking host if connected then will return mlflow, if any exception occur then NullTracker.

- In both the cases estimator will execute.

datasetinsights/estimators/faster_rcnn.py
86sanj: Poping tracking server details from clone of config so that only hyperparameter details can be logged to mlflow. 
- Original config will be as it is and if its required anywhere in code that can be used.

datasetinsights/io/tracker/factory.py
86sanj: - We need to return some instance on which logging can be handle. 
- With NullTracker class name " __getattr__" does not get called and to handle any method call. ex start_run.

datasetinsights/io/tracker/factory.py
86sanj: why override the default name?
- If user pass run name from config then that should be the part of run else we need to disable the option to pass run name through the config.

why isn't the name of the run updated through the tracker itself?
why does a static method in the factory take in params that look specific to mlflow?
- Because I was starting run at factory and you need run name during start. 
- I will move this code to tracker and will start run there.

datasetinsights/io/tracker/mlflow.py
86sanj: - If this run is executed by unity employee then client id is required, in this case it will throw exception and NullTracker will be initiated.
- If this run is executed by non unity employ, it will try to connect to given tracking host if connected then will return mlflow, if any exception occur then NullTracker.

- Do you want MLflow Tracking feature only for unity employee?

datasetinsights/io/tracker/mlflow.py
86sanj: I checked with Monica and Jash both. They don't have any idea about this. So I need to figure out and implement the changes. I check their setup doc and IAP PR, it does not have any setting for timeout. I need to explore google api/IAP setup doc and implement this. 

- Is this fine I do timeout increase change after merging this code?

datasetinsights/io/tracker/factory.py
adason: Did you successful make reference to NullTracker class without creating an instance?

adason (approved):

datasetinsights/io/tracker/factory.py
86sanj: With reference  _ _ getattr_ _(name) does not get called on any method call( ex: NullTracker.start_run()), so we need to create instance of NullTracker.

romerocesar (approved):

PR #127, opened 2020-11-16
Rename metrics for kubeflow UI

masonrubenstein:
# Peer Review Information

Kubeflow metrics must meet the regex restrictions to be shown on the UI/dashboard.

Source:
https://www.kubeflow.org/docs/pipelines/sdk/pipelines-metrics/
> `name` must follow the pattern `^[a-zA-Z]([-_a-zA-Z0-9]{0,62}[a-zA-Z0-9])?$`.
>
> For Kubeflow Pipelines version 0.5.1 or earlier, name must match the following pattern `^[a-z]([-a-z0-9]{0,62}[a-z0-9])?$`

# Pull Request Check List

### DIFF
--- a/datasetinsights/configs/faster_rcnn.yaml
+++ b/datasetinsights/configs/faster_rcnn.yaml
@@ -27,13 +27,13 @@ test:
       version: 2017
       split: val
 metrics:
-  mAP:
+  mean-average-precision:
     name: MeanAveragePrecisionAverageOverIOU
-  mAPIOU50:
+  mean-average-precision-iou50:
     name: MeanAveragePrecisionIOU50
-  mAR:
+  mean-average-recall:
     name: MeanAverageRecallAverageOverIOU
-  APIOU50:
+  average-precision-iou50:
     name: AveragePrecisionIOU50
 optimizer:
   name: Adam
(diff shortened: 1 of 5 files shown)


adason:
The pattern for kubeflow pipeline >0.5.1 was `^[a-z]([-a-z0-9]{0,62}[a-z0-9])?$` which is actually compatible with our current metric name such as `mAP`, `mAPIOU50`, etc. We might have very ancient version which used the pattern `^[a-z]([-a-z0-9]{0,62}[a-z0-9])?$` that will fail to parse those strings we currently have. This might call to upgrade our cluster at some point. 

Anyway, I do like the proposed change on the metric name as they are more explicit than vary mysterious name such as mAP.

adason:
ping @Saurav-D @BlairLee for awareness

adason (approved):

masonrubenstein:
@adason Agreed. We should have a policy on updating kubeflow or kubeflow/pipelines that aligns with changes like this or on some time schedule (every 6mo) due to the quick iteration nature of the repo. 

I can look into this if I can speak with someone who did the original setup to save some time on figuring out how much of a custom deployment we operate on. For ref: https://www.kubeflow.org/docs/gke/pipelines/upgrade/

Issue #54, opened 2020-08-21
Adding support for extra arguments to override the estimator yaml config in a kubeflow pipeline


AdamPalmarUnity:
/kind feature

**Why you need this feature:**

In order to avoid rebuilding the docker container to change the estimator yaml configuration. Allowing the user to provide extra arguments with the cli command for evaluate and train.  

**Describe the solution you'd like:**
For the estimator configuration 

```
estimator: FasterRCNN
backbone: resnet50
num_classes: 64
task: object_detection
train:
  dataset:
    name: SynDetection2D
    args:
      run_execution_id: zMvGEV0
      version: 0.0.1
      def_id: c31620e3-55ff-4af6-ae86-884aa0daa9b2
      random_seed: 47
      split: train
  batch_size: 4
  accumulation_steps: 8
  epochs: 50
  log_frequency: 10
val:
  dataset:
    name: SynDetection2D
    args:
      run_execution_id: zMvGEV0
      version: 0.0.1
      def_id: c31620e3-55ff-4af6-ae86-884aa0daa9b2
      random_seed: 47
      split: val
  batch_size: 4
test:
  dataset:
    name: GroceriesReal
    args:
      version: v3
      split: test
  batch_size: 4
metrics:
  mAP:
    name: MeanAveragePrecisionAverageOverIOU
  mAPIOU50:
    name: MeanAveragePrecisionIOU50
  mAR:
    name: MeanAverageRecallAverageOverIOU
  APIOU50:
    name: AveragePrecisionIOU50
optimizer:
  name: Adam
  args:
    lr: 2.0e-4
pretrained: False
pretrained_backbone: True
val_interval: 1
```

Being able to override the estimator config the following way.

The example below will override the optimizer.args.lr value and the number of train.epochs. 
`datasetinsights train --config=path/to/config --train-data=path optimizer.args.lr 0.00005 train.epochs 100   `

This will design will allow the cli to take extra arguments that will override the values in the config file. 

The proposed code change to the cli is the evaluate command as an example:

```

def cli(
    config,
    checkpoint_file,
    test_data,
    tb_log_dir,
    workers,
    kfp_metrics_dir,
    kfp_metrics_filename,
    no_cuda,
):
    ctx = click.get_current_context()
    logger.debug(f"Called evaluate command with parameters: {ctx.params}")
    logger.debug(f"Override estimator config with args: {ctx.args}")
    config = CN.load_cfg(open(config, "r"))

   ------ new code --------
[...]

PR #130, opened 2020-12-08
load config from gcs and local

### LINKED ISSUE #104: Read config yaml from remote (GCS/HTTP)
**Why you need this feature:**
If we want to change the config, we have to build the docker image again and use that custom image.

**Describe the solution you'd like:**
If we can read config from remote location like gcs or http then we don't have to use custom docker image.

86sanj:
# Peer Review Information
1. Load config yaml from remote(HTTP(S)/GCS) and local location with absolute and relative path
2.  docs/.rst file
3.  Unit test case

# Pull Request Check List

### DIFF
--- a/datasetinsights/commands/evaluate.py
+++ b/datasetinsights/commands/evaluate.py
@@ -4,7 +4,7 @@ import click
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
-from datasetinsights.io.config_handler import handle_config
+from datasetinsights.io.config_handler import prepare_config
 
 logger = logging.getLogger(__name__)
 
@@ -85,10 +85,10 @@ logger = logging.getLogger(__name__)
     ),
 )
 @click.option(
-    "--access-token",
+    "--override",
     type=str,
     default=None,
-    help="access token to authenticate remote YAML download. ",
+    help="key-value pairs to override config. ",
 )
 def cli(
     config,
@@ -100,12 +100,12 @@ def cli(
     kfp_metrics_filename,
     kfp_ui_metadata_filename,
     no_cuda,
-    access_token,
+    override,
 ):
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    config = handle_config(path=config, access_token=access_token)
+    config = prepare_config(path=config, override=override)
     estimator = create_estimator(
         config=config,
         name=config.estimator,
--- a/datasetinsights/commands/evaluate.py
+++ b/datasetinsights/commands/evaluate.py
@@ -4,7 +4,7 @@ import click
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
-from datasetinsights.io.config_handler import load_config
+from datasetinsights.io.config_handler import handle_config
 
 logger = logging.getLogger(__name__)
 
@@ -84,6 +84,12 @@ logger = logging.getLogger(__name__)
         "model will be trained using CUDA."
     ),
 )
+@click.option(
+    "--access-token",
+    type=str,
+    default=None,
+    help="access token to authenticate remote YAML download. ",
+)
 def cli(
     config,
     checkpoint_file,
@@ -94,11 +100,12 @@ def cli(
     kfp_metrics_filename,
     kfp_ui_metadata_filename,
     no_cuda,
+    access_token,
 ):
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    config = load_config(config)
+    config = handle_config(path=config, access_token=access_token)
     estimator = create_estimator(
         config=config,
         name=config.estimator,
--- a/datasetinsights/commands/evaluate.py
+++ b/datasetinsights/commands/evaluate.py
@@ -4,7 +4,7 @@ import click
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
-from datasetinsights.io.config_handler import ConfigHandler
+from datasetinsights.io.config_handler import load_config
 
 logger = logging.getLogger(__name__)
 
@@ -98,7 +98,7 @@ def cli(
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    config = ConfigHandler.load_config(config)
+    config = load_config(config)
     estimator = create_estimator(
         config=config,
         name=config.estimator,
--- a/datasetinsights/commands/evaluate.py
+++ b/datasetinsights/commands/evaluate.py
@@ -4,7 +4,7 @@ import click
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
-from datasetinsights.io.config_handler import prepare_config
+from datasetinsights.io.config_handler import load_config
 
 logger = logging.getLogger(__name__)
 
@@ -84,12 +84,6 @@ logger = logging.getLogger(__name__)
         "model will be trained using CUDA."
     ),
 )
-@click.option(
-    "--override",
-    type=str,
-    default=None,
-    help="key-value pairs to override config. ",
-)
 def cli(
     config,
     checkpoint_file,
@@ -100,12 +94,11 @@ def cli(
     kfp_metrics_filename,
     kfp_ui_metadata_filename,
     no_cuda,
-    override,
 ):
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    config = prepare_config(path=config, override=override)
+    config = load_config(config)
     estimator = create_estimator(
         config=config,
         name=config.estimator,
--- a/datasetinsights/commands/evaluate.py
+++ b/datasetinsights/commands/evaluate.py
@@ -1,10 +1,10 @@
 import logging
 
 import click
-from yacs.config import CfgNode as CN
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
+from datasetinsights.io.config_handler import ConfigHandler
 
 logger = logging.getLogger(__name__)
 
@@ -98,7 +98,7 @@ def cli(
     ctx = click.get_current_context()
     logger.debug(f"Called evaluate command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
-    config = CN.load_cfg(open(config, "r"))
+    config = ConfigHandler.load_config(config)
     estimator = create_estimator(
         config=config,
         name=config.estimator,
--- a/datasetinsights/io/config_handler.py
+++ b/datasetinsights/io/config_handler.py
@@ -1,0 +1,36 @@
+import logging
+import os
+import tempfile
+
+from yacs.config import CfgNode
+
+from datasetinsights.constants import GCS_BASE_STR
+from datasetinsights.io import create_downloader
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigHandler:
+    """ Handles all the config related tasks such as
+     loading config from local or remote locations.
+    """
+
+    @staticmethod
+    def load_config(config):
+        """load config from local or remote location .
+
+        Args:
+            config: config location
+        Returns:
+            loaded config object of type dictionary
+        """
+        if config.startswith(GCS_BASE_STR):
+            downloader = create_downloader(source_uri=config)
+            with tempfile.TemporaryDirectory() as tmp:
+                downloader.download(source_uri=config, output=tmp)
+                logger.info(f"downloading to directory: {tmp}")
+                file_path = os.path.join(tmp, config.split("/")[-1])
+                logger.info(f"loaded config from {config}")
+                return CfgNode.load_cfg(open(file_path, "r"))
+        logger.info(f"loaded config from {config}")
+        return CfgNode.load_cfg(open(config, "r"))
--- a/datasetinsights/io/config_handler.py
+++ b/datasetinsights/io/config_handler.py
@@ -1,36 +1,38 @@
 import logging
-import os
 import tempfile
+from pathlib import Path
 
 from yacs.config import CfgNode
 
-from datasetinsights.constants import GCS_BASE_STR
+import datasetinsights.constants as const
 from datasetinsights.io import create_downloader
 
 logger = logging.getLogger(__name__)
 
 
-class ConfigHandler:
-    """ Handles all the config related tasks such as
-     loading config from local or remote locations.
+def load_config(path):
+    """load config from local or remote location .
+
+    Args:
+        path: config location
+    Examples:
+         path="gs://thea-dev/../config.yaml"
+         path="https://thea-dev/../config.yaml"
+         path="http://thea-dev/../config.yaml"
+         path="/root/../config.yaml"
+    Returns:
+        loaded config object of type dictionary
     """
-
-    @staticmethod
-    def load_config(config):
-        """load config from local or remote location .
-
-        Args:
-            config: config location
-        Returns:
-            loaded config object of type dictionary
-        """
-        if config.startswith(GCS_BASE_STR):
-            downloader = create_downloader(source_uri=config)
-            with tempfile.TemporaryDirectory() as tmp:
-                downloader.download(source_uri=config, output=tmp)
-                logger.info(f"downloading to directory: {tmp}")
-                file_path = os.path.join(tmp, config.split("/")[-1])
-                logger.info(f"loaded config from {config}")
-                return CfgNode.load_cfg(open(file_path, "r"))
-        logger.info(f"loaded config from {config}")
-        return CfgNode.load_cfg(open(config, "r"))
+    if path.startswith(
+        (const.GCS_BASE_STR, const.HTTP_URL_BASE_STR, const.HTTPS_URL_BASE_STR,)
+    ):
+        downloader = create_downloader(source_uri=path)
+        with tempfile.TemporaryDirectory() as tmp:
+            downloader.download(source_uri=path, output=tmp)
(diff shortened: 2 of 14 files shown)


datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,36 @@
...
+
+from yacs.config import CfgNode
+
+from datasetinsights.constants import GCS_BASE_STR
+from datasetinsights.io import create_downloader
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigHandler:
```
adason: Why does this needs to be a python class? Can this method be a simple python function?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,36 @@
...
+    @staticmethod
+    def load_config(config):
+        """load config from local or remote location .
+
+        Args:
+            config: config location
+        Returns:
+            loaded config object of type dictionary
+        """
+        if config.startswith(GCS_BASE_STR):
```
adason: What if the config path starts with `http/https`? What if it starts with `/path/to/file`? What are different edge cases? What's the expected behavior if the file fail to download?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,36 @@
...
+            downloader = create_downloader(source_uri=config)
+            with tempfile.TemporaryDirectory() as tmp:
+                downloader.download(source_uri=config, output=tmp)
+                logger.info(f"downloading to directory: {tmp}")
+                file_path = os.path.join(tmp, config.split("/")[-1])
```
adason: [nit] It might be easier to handle file path using `pathlib` in standard library.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,36 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ConfigHandler:
+    """ Handles all the config related tasks such as
+     loading config from local or remote locations.
+    """
+
+    @staticmethod
+    def load_config(config):
```
adason: I think it's better to rename variable here to `def load_config(path)`. Also the docstring should mention what type of paths are supported. Bonus point if some examples are provided. 

Did you update the corresponding files in `docs/` so that the docstring can be updated properly?

adason (changes requested):

adason:
https://github.com/Unity-Technologies/datasetinsights/issues/104

datasetinsights/io/config_handler.py
86sanj: In this issue - https://github.com/Unity-Technologies/datasetinsights/issues/54 it was mentioned only GCS so I have implemented for GCS but now I have changed to handle HTTP/HTTPS also.

Supported locations:
1. GCS
2. HTTP/HTTPS
3. Local file: /a/cofig.yam

- If downloading or loading fails, run should terminate with caused exception.
- For HTTP/HTTPS:  Do you want to support only open URL or login based URL also? If yes then we need to add access token in train and evaluate cli.

datasetinsights/io/config_handler.py
86sanj: I will update.  I have missed to update for Mlflow change also, I will include in this commit.

datasetinsights/commands/evaluate.py
```diff
@@ -84,6 +84,12 @@
         "model will be trained using CUDA."
     ),
 )
+@click.option(
```
adason: not clear what's the access token here and what should be provided in this case. If it's GCS path, the token should came from environment variable (GCS best practice). If it's http path, we should assume it's publicly available. It's very hard to guess what's the auth requirements from http url.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+
+from yacs.config import CfgNode
+
+import datasetinsights.constants as const
+from datasetinsights.io import create_downloader
+
+logger = logging.getLogger(__name__)
+
+
+def handle_config(path=None, access_token=None):
```
adason: The name `handle_config` is very vague. It should be explicit, such as `load` or `load_config`.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    """load and override config from local or remote location .
```
adason: Why does this method need to override config?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        path: config location
+        access_token: authenticate remote location
+    Examples:
+         path="gs://thea-dev/../config.yaml"
+         path="https://thea-dev/../config.yaml"
+         path="http://thea-dev/../config.yaml"
+         path="/root/../config.yaml"
```
adason: [nit] Would it be better if we explicitly require use to use path like `file://path/to/config.yaml` with the `file://` prefix so that it's clear this path is expected to be local filesystem?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    Returns:
+        config object of type yacs.config.CfgNode
+    """
+    if path.startswith(
```
adason: What happened if the prefix was not recognized?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+        (const.GCS_BASE_STR, const.HTTP_URL_BASE_STR, const.HTTPS_URL_BASE_STR,)
```
adason: What if we need to support other path in the future (say `s3://...`)? This check will need to be changed in-line here. Can this be stored as constant? Can this be shared with the downloader factory? Can you let `create_downloader` handle cases when the url is not recognized?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    ):
+        downloader = create_downloader(
+            source_uri=path, access_token=access_token
+        )
+        with tempfile.TemporaryDirectory() as tmp:
+            downloader.download(source_uri=path, output=tmp)
+            logger.info(f"downloading to directory: {tmp}")
```
adason: use debug logging instead of info

tests/test_config_handler.py
```diff
@@ -0,0 +1,71 @@
+from unittest.mock import MagicMock, patch
+
+from datasetinsights.io.config_handler import handle_config
+
+
```
adason, line 7: Can you add test cases when the path is invalid (or not supported)

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    Args:
+        path: config location
+        access_token: authenticate remote location
+    Examples:
+         path="gs://thea-dev/../config.yaml"
+         path="https://thea-dev/../config.yaml"
+         path="http://thea-dev/../config.yaml"
+         path="/root/../config.yaml"
+    Returns:
+        config object of type yacs.config.CfgNode
```
adason, line 34: Can you make sure this docstring is parsed properly in sphinx?

adason (changes requested):

datasetinsights/io/config_handler.py
86sanj: Is this ok if we have following URL to distinguish between absolute and relative local path. 
Ex:
- file://datasetinsights/configs/faster_rcnn_groceries_real.yaml - relative path
- file:///user/a/b/datasetinsights/configs/faster_rcnn_groceries_real.yaml - absolute path

datasetinsights/io/config_handler.py
86sanj: Raise ValueError with message:"Prefix does not support for given path"

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,86 @@
...
+    if not path.startswith(const.LOCAL_FILE_BASE_STR):
+        try:
+            downloader = create_downloader(source_uri=path)
+            with tempfile.TemporaryDirectory() as tmp:
+                downloader.download(source_uri=path, output=tmp)
+                logger.debug(f"downloading to directory: {tmp}")
+                config_path = Path(path)
+                file_path = tmp / config_path.relative_to(config_path.parent)
+                return CfgNode.load_cfg(open(file_path, "r"))
+        except ValueError:
```
adason: From the exception message, `ValueError` was specifically raised from `create_downloader`. ValueError is actually a very common error in python that can be raised from different modules. Can you add try-catch for `create_downloader` only? 

Also, it seems create_downloader will raise exceptions with proper error message. Why not just let it raise?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,86 @@
...
+            logger.exception(f"Prefix does not support for path: {path}")
+            raise ValueError
+        except Exception:
```
adason: You should catch this exception error message and send it to logs here. It should be helpful to let user know what's the root cause of exceptions rather than hiding it. For example:
```
except Exception as e: 
    logger.execption("some message")
    logger.exception(e)
    raise  
```

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,86 @@
...
+    try:
+        logger.debug(f" config before overriding {config}")
+        tokens = override.split()
+        merge_list = []
+        for token in tokens:
+            merge_list.extend(token.split(_EQUAL_STR))
+        logger.info(f" overriding key-values {merge_list}")
+        config.merge_from_list(merge_list)
+        logger.info(f" overriding completed, config after override{config}")
+    except AssertionError:
```
adason: It's not clear when AssertionError is raised. It's usually easier to have 1 to 2 lines of code in the try clause so that it's easier to handle know errors. I will encourage you to read item13 of the effective python book.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,86 @@
...
+    Args:
+        override: string of key-value pairs
+        config: config object of type yacs.config.CfgNode
+    Examples:
+         override="optimizer.args.lr=0.00005 pretrained=False"
+    """
+
+    try:
+        logger.debug(f" config before overriding {config}")
+        tokens = override.split()
```
adason: I will recommend doing some validation against the override string to catch issues early, instead of relying on `CfgNode` to catch errors. Using regexp is usually very helpful for strings. If the override string does not match the expected pattern, it should raise immediately.

adason (changes requested):

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,87 @@
...
+    Examples:
+         override="optimizer.args.lr=0.00005 pretrained=False"
+    """
+
+    pattern = re.compile(_REGEX_PTRN)
+    logger.debug(f" config before overriding {config}")
+    tokens = override.split()
+    merge_list = []
+    for token in tokens:
+        if pattern.match(token):
```
adason: Why use regex pattern match on each token after split? You should be able to validate the whole regex input string directly. 

https://github.com/Unity-Technologies/datasetinsights/blob/master/datasetinsights/commands/download.py#L12

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,87 @@
...
+            with tempfile.TemporaryDirectory() as tmp:
+                downloader.download(source_uri=path, output=tmp)
+                logger.debug(f"downloading to directory: {tmp}")
+                config_path = Path(path)
+                file_path = tmp / config_path.relative_to(config_path.parent)
+                return CfgNode.load_cfg(open(file_path, "r"))
+        except Exception as e:
+            logger.exception(f"Config loading from location{path} failed")
+            logger.exception(e)
+            raise e
```
adason: [nit] Don't need to raise with e. You can simply call `raise` without passing exception objects e.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def load_config(path):
+    """ load config from local or remote locations .
+    Examples:
+        >>> path="gs://thea-dev/../config.yaml"
+        >>> path="https://thea-dev/../config.yaml"
+        >>> path="http://thea-dev/../config.yaml"
+        >>> path="file:///root/../config.yaml" # absolute path
+        >>> path="file://datasetinsights/../config.yaml" # relative local path
```
adason: Let's get rid of supports for relative local path. It's very error prone to use relative local path.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,87 @@
...
+        try:
+            downloader = create_downloader(source_uri=path)
+            with tempfile.TemporaryDirectory() as tmp:
+                downloader.download(source_uri=path, output=tmp)
+                logger.debug(f"downloading to directory: {tmp}")
+                config_path = Path(path)
+                file_path = tmp / config_path.relative_to(config_path.parent)
+                return CfgNode.load_cfg(open(file_path, "r"))
+        except Exception as e:
+            logger.exception(f"Config loading from location{path} failed")
```
adason: If the users use `s3://...` as the config path, they will see this error message, but it's not clear why this failed. I think we should validate against supported config path prefix, as in https://github.com/Unity-Technologies/datasetinsights/blob/master/datasetinsights/commands/download.py#L12

adason (changes requested):

kubeflow/compiled/evaluate_the_model.yaml
```diff
@@ -136,7 +136,7 @@ spec:
     parameters:
     - {name: docker, value: 'unitytechnologies/datasetinsights:latest'}
     - {name: source_uri, value: 'https://storage.googleapis.com/datasetinsights/data/groceries/v3.zip'}
-    - {name: config, value: datasetinsights/configs/faster_rcnn_groceries_real.yaml}
+    - {name: config, value: 'file://datasetinsights/configs/faster_rcnn_groceries_real.yaml'}
```
adason: Can you change the default value to absolute path. We discussed dropping support for relative path.

tests/test_config_handler.py
```diff
@@ -0,0 +1,72 @@
...
+@patch("datasetinsights.io.config_handler.create_downloader")
+@patch("datasetinsights.io.config_handler.CfgNode.load_cfg")
+@patch("datasetinsights.io.config_handler.open")
+def test_http_load_config(mock_open, mock_config, mock_create_downloader):
+    mock_obj = MagicMock()
+    mock_create_downloader.return_value = mock_obj
+    mock_open.return_value = mock_obj
+
+    config_url = "http://thea-dev/../config.yaml"
```
adason: Why not just use concrete url such as `http://www.file.com/config.yaml`? Why test against an unrealistic url with `..` in there?

tests/test_config_handler.py
```diff
@@ -0,0 +1,72 @@
...
+    mock_obj.download.assert_called_once()
+
+
+@patch("datasetinsights.io.config_handler.CfgNode.load_cfg")
+@patch("datasetinsights.io.config_handler.open")
+def test_local_load_config(mock_open, mock_config):
+    mock_obj = MagicMock()
+    mock_open.return_value = mock_obj
+
+    config_url = "file://test.yaml"
```
adason: Please drop support for relative path.

tests/test_config_handler.py
```diff
@@ -0,0 +1,72 @@
...
+    load_config(config_url)
+    mock_config.assert_called_with(mock_obj)
+
+
+@patch("datasetinsights.io.config_handler.create_downloader")
```
adason: Why is create_downloader mocked in this case? Is the mocking necessary, if the bad_url will fail the load_config immediately?

adason (dismissed):

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,40 @@
...
+from yacs.config import CfgNode
+
+import datasetinsights.constants as const
+from datasetinsights.io import create_downloader
+
+logger = logging.getLogger(__name__)
+_REMOTE_PATTERN = r"://"
+
+
+def load_config(path):
```
adason, line 19: Improve the docstring here.

kubeflow/compiled/evaluate_the_model.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: argoproj.io/v1alpha1
 kind: Workflow
 metadata:
   generateName: evaluate-the-model-
-  annotations: {pipelines.kubeflow.org/kfp_sdk_version: 1.0.1, pipelines.kubeflow.org/pipeline_compilation_time: '2020-11-04T17:16:55.837308',
+  annotations: {pipelines.kubeflow.org/kfp_sdk_version: 1.0.1, pipelines.kubeflow.org/pipeline_compilation_time: '2021-01-13T16:53:54.256807',
```
adason: Why do you need to re-compile the pipeline if nothing has changed?

kubeflow/compiled/train_and_evaluate_combined_pipeline.yaml
```diff
@@ -135,7 +135,7 @@ spec:
     container:
       args: ['--config={{inputs.parameters.train_config}}', --train-data=/data, --val-data=/data,
         '--tb-log-dir={{inputs.parameters.tb_log_dir}}', --kfp-log-dir=/kfp_logs,
-        '--kfp-ui-metadata-filename=kfp_ui_metadata.json--checkpoint-dir={{inputs.parameters.checkpoint_dir}}']
+        --kfp-ui-metadata-filename=kfp_ui_metadata.json, '--checkpoint-dir={{inputs.parameters.checkpoint_dir}}']
```
adason, line 138: Great catch!

tests/test_config_handler.py
```diff
@@ -0,0 +1,70 @@
...
+@patch("datasetinsights.io.config_handler.CfgNode.load_cfg")
+@patch("datasetinsights.io.config_handler.open")
+def test_local_load_config(mock_open, mock_config):
+    mock_obj = MagicMock()
+    mock_open.return_value = mock_obj
+
+    config_url = "file:///root/test.yaml"
+    load_config(config_url)
+    mock_config.assert_called_with(mock_obj)
+
```
adason, line 64: Can you add test cases for 

```
path="/root/config.yaml" # absolute path
path="datasetinsights/config.yaml" # relative path
```

adason (changes requested):

adason (approved):

PR #133, opened 2021-01-05
MLFlow: Lookup client_id and host_id from Kubernetes env variable

86sanj:
# Peer Review Information
1. Lookup mlflow client_id and host_id from Kubernetes env variable, if both are configured in config YAML then override with YAML values.
2. Updated pipelines.py with env variable lookup, compiled pipelines
3. Unit test case

Kubeflow run_id: train_on_real_world_dataset-test 5
# Pull Request Check List

### DIFF
--- a/datasetinsights/io/tracker/factory.py
+++ b/datasetinsights/io/tracker/factory.py
@@ -48,23 +48,19 @@ class TrackerFactory:
             raise InvalidTrackerError
 
     @staticmethod
-    def _mlflow_tracker_instance(mlflow_config):
+    def _mlflow_tracker_instance(config):
 
         """Static instance access method.
 
         Args:
-            host_id: MlTracker server host
-            client_id: MLFlow tracking server client id
-            exp_name: name of the experiment
+            config : config object, holds server details
         Returns:
             tracker singleton instance.
         """
         if not TrackerFactory.__tracker_instance:
             with TrackerFactory.__singleton_lock:
                 if not TrackerFactory.__tracker_instance:
-                    TrackerFactory.__tracker_instance = MLFlowTracker(
-                        mlflow_config
-                    )
+                    TrackerFactory.__tracker_instance = MLFlowTracker(config)
         logger.info("getting tracker instance")
         return TrackerFactory.__tracker_instance
 
--- a/datasetinsights/io/tracker/factory.py
+++ b/datasetinsights/io/tracker/factory.py
@@ -56,7 +56,7 @@ class TrackerFactory:
             raise InvalidTrackerError
 
     @staticmethod
-    def _mlflow_tracker_instance(**kwarg):
+    def _mlflow_tracker_instance(**kwargs):
 
         """Static instance access method.
 
@@ -68,7 +68,7 @@ class TrackerFactory:
         if not TrackerFactory.__tracker_instance:
             with TrackerFactory.__singleton_lock:
                 if not TrackerFactory.__tracker_instance:
-                    TrackerFactory.__tracker_instance = MLFlowTracker(**kwarg)
+                    TrackerFactory.__tracker_instance = MLFlowTracker(**kwargs)
         logger.info("getting tracker instance")
         return TrackerFactory.__tracker_instance
 
--- a/datasetinsights/io/tracker/factory.py
+++ b/datasetinsights/io/tracker/factory.py
@@ -69,7 +69,7 @@ class TrackerFactory:
         """Static instance access method.
 
         Args:
-            kwarg : key-value pairs of mlflow parameters
+            kwargs : key-value pairs of mlflow parameters
         Returns:
             tracker singleton instance.
         """
--- a/datasetinsights/io/tracker/factory.py
+++ b/datasetinsights/io/tracker/factory.py
@@ -31,9 +31,17 @@ class TrackerFactory:
         if TrackerFactory.MLFLOW_TRACKER == tracker_type:
 
             try:
-                mlf_tracker = TrackerFactory._mlflow_tracker_instance(
-                    config
-                ).get_mlflow()
+                tracker = config.get(TrackerFactory.TRACKER, None)
+                if tracker and tracker.get(TrackerFactory.MLFLOW_TRACKER, None):
+                    mlflow_config = tracker.get(TrackerFactory.MLFLOW_TRACKER)
+
+                    mlf_tracker = TrackerFactory._mlflow_tracker_instance(
+                        **mlflow_config
+                    ).get_mlflow()
+                else:
+                    mlf_tracker = (
+                        TrackerFactory._mlflow_tracker_instance().get_mlflow()
+                    )
                 logger.info("initializing mlflow_tracker")
                 return mlf_tracker
             except Exception as e:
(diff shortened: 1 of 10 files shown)


datasetinsights/io/tracker/mlflow.py
```diff
@@ -45,28 +45,27 @@ class MLFlowTracker:
     EXP_NAME = "experiment"
     RUN_NAME = "run"
     DEFAULT_RUN_NAME = "run-" + TIMESTAMP_SUFFIX
+    TRACKER = "tracker"
+    MLFLOW_TRACKER = "mlflow"
+    DEFAULT_EXP_NAME = "datasetinsights"
 
-    def __init__(self, mlflow_config):
+    def __init__(self, config):
```
adason: [mustfix]

We missed this in the first review. I think it's actually more robust to have host_id, client_id, exp_name and run_name to be the variable passed to the tracker instnace. I would expect the code to be called like this:


```
# factory.py

mlflow_config = config["trackers"]["mlflow"]
mlf_tracker = TrackerFactory.create(**mlflow_config)

# mlflow.py

__init__(self, *, client_id=None, host_id=None, run_name=None, exp_name=DEFAULT_EXP_NAME):
...
```

in this case, python will unpack all values for you.

adason: This is more pythontic and does not require you to write code such as `mlflow_config.get(MLFlowTracker.HOST_ID)` in order to fetch variables stored in the config dict. It's also easier to maintain as users know what to expect in order to properly initialized an mlflow tracker instance, without looking onto the code itself. Also, how each variable is used, whether it's required or not should be part of the docstring.

datasetinsights/io/tracker/mlflow.py
```diff
@@ -45,31 +45,40 @@ class MLFlowTracker:
...
+    MLFLOW_TRACKER = "mlflow"
+    DEFAULT_EXP_NAME = "datasetinsights"
 
-    def __init__(self, mlflow_config):
+    def __init__(
+        self,
+        *,
+        client_id=None,
+        host=None,
+        run=DEFAULT_RUN_NAME,
```
adason: Why not call this `run_name`?

datasetinsights/io/tracker/factory.py
```diff
@@ -55,23 +56,19 @@ def create(config=None, tracker_type=None):
             raise InvalidTrackerError
 
     @staticmethod
-    def _mlflow_tracker_instance(mlflow_config):
+    def _mlflow_tracker_instance(**kwarg):
```
adason: [nit] Use `**kwargs` instead. This is a variable commonly used in python packages.

datasetinsights/io/tracker/mlflow.py
```diff
@@ -80,6 +89,42 @@ def get_mlflow(self):
         logger.info("get mlflow")
         return self.__mlflow
 
+    @staticmethod
+    def _get_variables(client_id=None, host_id=None):
```
adason: Why not just split into two methods: `_get_client_id(...)` and `_get_host_id(...)`.  Each method should do one and only one thing. 


Also, it might be easier to use default value pattern like:

```
def __init__(self, clinet_id=None, host_id=None,...):
  client_id = client_id | self._get_client_id(...)
  host_id = host_id | self._get_host_id(...)
  ....
```

This is python's syntax sugar set default value when the variable is None.

datasetinsights/io/tracker/mlflow.py
```diff
@@ -45,31 +45,40 @@ class MLFlowTracker:
...
+    TRACKER = "tracker"
+    MLFLOW_TRACKER = "mlflow"
+    DEFAULT_EXP_NAME = "datasetinsights"
 
-    def __init__(self, mlflow_config):
+    def __init__(
+        self,
+        *,
+        client_id=None,
+        host=None,
```
adason, line 70: Why not call this `host_id` to make it consistent across the codebase?

datasetinsights/io/tracker/mlflow.py
86sanj: Because  **args map dictionary with key name to the method argument, and during MLFlow design we decided following names for mlflow config variables:
 mlflow:
   experiment:
   run:
   client_id: 
   host:

datasetinsights/io/tracker/mlflow.py
86sanj: Same as above!

datasetinsights/io/tracker/mlflow.py
86sanj: For end users who do not have mlflow setup, for them both values around "|" will be None, so it will throw below error.
-           host_id = host | os.environ.get("MLFLOW_HOST_ID", None)
-          TypeError: unsupported operand type(s) for |: 'NoneType' and 'NoneType'

datasetinsights/io/tracker/mlflow.py
adason: In this case we should make variable naming consistent.

datasetinsights/io/tracker/mlflow.py
adason: In this case host_id is required. Why do you need default value to be None? If host_id is not provided, the initialization method should fail immediately when the constructor is called.

86sanj: - host_id is required only if you have mlflow setup. 
- In case of public users who do not have mlflow setup for them host_id will be None in constructor, then it will lookup in env variable and there also host_id won't be configured so if default value lookup won't be None then it will throw key error so fetching hosted_id by putting default None: ' os.environ.get("MLFLOW_HOST_ID", None)'.

adason: >host_id is required only if you have mlflow setup

If this is the case, why will `MLFlowTracker` fail to initialize if both client_id and host_id are `None`? 

> In case of public users who do not have mlflow setup for them host_id will be None in constructor, then it will lookup in env variable and there also host_id won't be configured so if default value lookup won't be None then it will throw key error so fetching hosted_id by putting default None: ' os.environ.get("MLFLOW_HOST_ID", None)'.

Let's document this in `MLFlowTracker` class docstring. It should be clear to users what's required.

86sanj: >If this is the case, why will MLFlowTracker fail to initialize if both client_id and host_id are None?


As mlflow is not setup and hence you don't have both Ids so fail to initialize and trigger Null tracker.

datasetinsights/io/tracker/factory.py
```diff
@@ -31,22 +32,30 @@ def create(config=None, tracker_type=None):
...
+            except ValueError as e:
+                logger.warning(
+                    "failed mlflow initialization, " "Host is not configured", e
+                )
+            except MlflowException as e:
+                logger.warning("failed mlflow initialization,", e)
+            except Exception as e:
+                logger.warning(
+                    "failed mlflow initialization, " "starting null_tracker", e
+                )
```
adason, line 58: [nit] Why not use `else` clause to initialize null tracker here in the try...except...else clause?

adason (dismissed):

adason:
@Saurav-D Can you also review this PR?

datasetinsights/io/tracker/factory.py
```diff
@@ -55,23 +64,19 @@ def create(config=None, tracker_type=None):
...
-    def _mlflow_tracker_instance(mlflow_config):
+    def _mlflow_tracker_instance(**kwargs):
 
         """Static instance access method.
 
         Args:
-            host_id: MlTracker server host
-            client_id: MLFlow tracking server client id
-            exp_name: name of the experiment
+            kwarg : key-value pairs of mlflow parameters
```
Saurav-D: [nit] shouldn't `kwargs` be the param name in docstring?

Saurav-D (dismissed):

datasetinsights/io/tracker/factory.py
86sanj: - If I put NullTracker under else block then when exception occurs Null tracker won't start and hence experiment run will terminate as there is no tracker for logging.
- We want to run experiment even mlflow fail to start so need to start NullTracker after except so that when exception occurs, log it and start the null tracker.

Saurav-D (approved):

PR #135, opened 2021-01-08
Cli config override

86sanj:
# Peer Review Information

1. Overriding config key-values from cli override command. 
2. Unit test cases

# Pull Request Check List

### DIFF
--- a/datasetinsights/commands/train.py
+++ b/datasetinsights/commands/train.py
@@ -1,12 +1,45 @@
 import logging
+import re
 
 import click
 
 import datasetinsights.constants as const
 from datasetinsights.estimators.base import create_estimator
-from datasetinsights.io.config_handler import load_config
+from datasetinsights.io.config_handler import prepare_config
 
 logger = logging.getLogger(__name__)
+
+
+class OverrideKey(click.ParamType):
+    """Source URI Parameter.
+
+    Args:
+        click ([type]): [description]
+
+    Raises:
+        click.BadParameter: [description]
+
+    Returns:
+        [type]: [description]
+    """
+
+    OVERRIDE_PTRN = (
+        r"^([:/\w_+-.%]+=[:/\w_+-.%]+)(\s[:/\w_+-.%]+=[:/\w_+-.%]+)*$"
+    )
+
+    def convert(self, value, param, ctx):
+        """ Validate override key and Converts the value.
+        """
+        pattern = re.compile(self.OVERRIDE_PTRN)
+        match = pattern.match(value)
+        if not match:
+            message = (
+                f"The override format {value} is not supported. "
+                f"Pattern: {self.OVERRIDE_PTRN}"
+            )
+            logger.exception(message)
+            self.fail(message, param, ctx)
+        return value
 
 
 @click.command(
@@ -105,6 +138,16 @@ logger = logging.getLogger(__name__)
     default=False,
     help="Force to disable validations.",
 )
+@click.option(
+    "--override",
+    default=None,
+    type=OverrideKey(),
+    required=False,
+    help=(
+        "String of key-value pairs."
+        f"Supported override key {OverrideKey.OVERRIDE_PTRN}"
+    ),
+)
 def cli(
     config,
     train_data,
@@ -117,12 +160,13 @@ def cli(
     workers,
     no_cuda,
     no_val,
+    override,
 ):
     ctx = click.get_current_context()
     logger.debug(f"Called train command with parameters: {ctx.params}")
     logger.debug(f"Override estimator config with args: {ctx.args}")
 
-    config = load_config(config)
+    config = prepare_config(path=config, override=override)
 
     estimator = create_estimator(
         name=config.estimator,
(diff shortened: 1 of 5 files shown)


datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+
+from datasetinsights.io.exceptions import InvalidOverrideKey
+
+logger = logging.getLogger(__name__)
+_REGEX_PTRN = r"^([:/\w_+-.%]+=[:/\w_+-.%]+)(\s[:/\w_+-.%]+=[:/\w_+-.%]+)*$"
+_EQUAL_STR = "="
+
+
+def override_config(override=None, config=None):
+    """override config params of config YAML, if override
```
adason: Please follow PEP257 to format multi-line docstrings. 

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    key doesn't exist it will throw Non-existent key error .
```
adason: Why not just explicitly say `InvalidOverrideKey` error?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    Args:
+        override: string of key-value pairs
+        config: config object of type yacs.config.CfgNode
+    Examples:
```
adason: This is not a fully example that can be run by [doctest](https://docs.python.org/3/library/doctest.html). Since we are writing examples here, why not provide better documented examples?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
+import logging
+import re
+
+from datasetinsights.io.exceptions import InvalidOverrideKey
+
+logger = logging.getLogger(__name__)
+_REGEX_PTRN = r"^([:/\w_+-.%]+=[:/\w_+-.%]+)(\s[:/\w_+-.%]+=[:/\w_+-.%]+)*$"
+_EQUAL_STR = "="
```
adason, line 13: _EQUAL_STR variable should be referenced in _REGEX_PTRN.

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
+import logging
+import re
+
+from datasetinsights.io.exceptions import InvalidOverrideKey
+
+logger = logging.getLogger(__name__)
+_REGEX_PTRN = r"^([:/\w_+-.%]+=[:/\w_+-.%]+)(\s[:/\w_+-.%]+=[:/\w_+-.%]+)*$"
```
adason: [nit] I suggest using explicit variable name like `_OVERRIDE_PATTHER`

tests/test_config_handler.py
```diff
@@ -0,0 +1,22 @@
...
+from datasetinsights.io.exceptions import InvalidOverrideKey
+
+
+@patch("yacs.config.CfgNode")
+def test_override_config(mock_cfg):
+
+    override = "a.b.c=test1 c.a_b=test2"
+    override_token = ["a.b.c", "test1", "c.a_b", "test2"]
+    override_config(override=override, config=mock_cfg)
+    mock_cfg.merge_from_list.assert_called_with(override_token)
```
adason: [nit] Why not just test against the overriden config that was returned by `override_config` method?

tests/test_config_handler.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+@patch("yacs.config.CfgNode")
```
adason: Why do you need mock config here? Why not just create real instance of CfgNode and pass to override_config here?

datasetinsights/io/config_handler.py
```diff
@@ -0,0 +1,41 @@
...
+    key doesn't exist it will throw Non-existent key error .
+
+    Args:
+        override: string of key-value pairs
+        config: config object of type yacs.config.CfgNode
+    Examples:
+         override="optimizer.args.lr=0.00005 pretrained=False"
+    """
+
+    pattern = re.compile(_REGEX_PTRN)
```
adason: Can the validation of override string separated from `override_config` method as it's own function?

datasetinsights/commands/train.py
```diff
@@ -105,6 +106,12 @@
     default=False,
     help="Force to disable validations.",
 )
+@click.option(
```
adason, line 141: It might be more convenient to have override string validation handled by click framework, similar to this:

https://github.com/Unity-Technologies/datasetinsights/blob/master/datasetinsights/commands/download.py#L12

adason (changes requested):

datasetinsights/io/config_handler.py
86sanj: I am using _EQUAL_STR to split override string and prepare list of key-value pairs which config.merge_from_list(merge_list) expect.

adason (dismissed):

adason (approved):

PR #140, opened 2021-01-27
Updated requirement lib and stable image

86sanj:
# Peer Review Information
This is to solve "Excess memory consumption error."
- https://readthedocs.org/projects/datasetinsights/builds/12853424/
- Existing Issue: https://github.com/readthedocs/readthedocs.org/issues/6692

Tested project on readthedoc: https://readthedocs.org/projects/readthedocs-memory-fix/builds/

# Pull Request Check List

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,6 +1,8 @@
 # .readthedocs.yml
 version: 2
 formats: all
+build:
+  image: stable
 python:
   version: 3.7
   install:


docs/requirements.txt
```diff
@@ -1,13 +1,17 @@
 absl-py==0.10.0
 alabaster==0.7.12
+alembic==1.4.1
```
BlairLee: Do we need to update those packages in order to fix the readthedocs build issue? How about we just add `image: stable`

86sanj: - As per docs/readme, we should have new lib added in requirement.txt, I was trying to add Mlflow dependency so added through given command "poetry export --dev --without-hashes -f requirements.txt > docs/requirements.txt". It added alot other dependency other than mlflow also.
- Let me try building without changing dependency.

86sanj: Ran 4 builds only on stable image, 2 passed 2 failed, not sure whether stable image will help or not. 
https://readthedocs.org/projects/readthedocs-memory-fix/builds/

adason (approved):

PR #141, opened 2021-01-28
Update synth training and fine-tune pipeline

BlairLee:
# Peer Review Information
Since 400K SynthDet V2 dataset requires 200G memory during training, I increase the memory limit in the synth training pipeline. 
I also add `train_split` in the fine-tune pipeline. Users can config which split they want to use during the fine-tuning. 
Here are some test runs on kubeflow: [fine-tune run](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/_/pipeline/#/runs/details/53b113ee-5456-4bcc-8250-d25bef7af9c4), [synth_train run](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/_/pipeline/#/runs/details/ea3450cb-7cd5-4669-af6f-8595825518cf).


kubeflow/pipelines.py
```diff
@@ -79,6 +79,7 @@ def train_op(
     num_gpu,
     gpu_type,
     checkpoint_file=None,
+    train_split=None,
```
adason: [mustfix] We should allow user to pass the whole override string so that they can override arbitrary config from the kubeflow UI.

kubeflow/pipelines.py
```diff
@@ -127,6 +128,8 @@ def train_op(
     ]
     if checkpoint_file:
         arguments.append(f"--checkpoint-file={checkpoint_file}")
+    if train_split:
```
adason: same as above.

kubeflow/pipelines.py
```diff
@@ -425,14 +430,15 @@ def train_on_synthetic_dataset_unity_simulation(
     config: str = "datasetinsights/configs/faster_rcnn_synthetic.yaml",
     tb_log_dir: str = "gs://<bucket>/runs/yyyymmdd-hhmm",
     checkpoint_dir: str = "gs://<bucket>/checkpoints/yyyymmdd-hhmm",
-    volume_size: str = "100Gi",
+    volume_size: str = "1.2Ti",
```
adason, line 451: Why change the default value here? This value can be changed by user when they are running kubeflow pipeline, right? Should we update the tutorial accordingly to let user know this value has to be increased according to their dataset size?

BlairLee: Agree. Updated the tutorial.

kubeflow/pipelines.py
BlairLee: Add override option in the pipeline. And tested in this [run](https://datasetinsights.endpoints.unity-ai-thea-test.cloud.goog/_/pipeline/#/runs/details/2ff13dcc-c1b2-4be8-b94d-0b28030a42c2).

kubeflow/pipelines.py
adason: I will suggest keeping default value so that user does not hit with huge resource budget. In the tutorial, it should be clear they need to increase this value if they are working on the dataset.

kubeflow/compiled/train_and_evaluate_combined_pipeline.yaml
```diff
@@ -80,6 +80,12 @@ spec:
         --kfp-ui-metadata-filename=kfp_ui_metadata.json, --kfp-metrics-filename=kfp_metrics.json]
       command: [datasetinsights, evaluate]
       env:
+      - name: MLFLOW_HOST_ID
```
adason: Please pull from master.

docs/source/Evaluation_Tutorial.md
```diff
@@ -46,7 +46,7 @@ You have to specify run parameters required by this pipeline:
 - `config`: Estimator config YAML file. You can use the default value which points to a YAML file packaged with our docker images or you can load from remote locations GCS or any HTTP(s) using file prefix `gs://, http(s)://`.
 - `tb_log_dir`: Path to store tensorboard logs used to visualize the training progress.
 - `checkpoint_dir`: Path to store output estimator checkpoints. These checkpoints represent your trained model and will be used for evaluation.
-- `volume_size`: Size of the Kubernetes Persistent Volume Claims (PVC) that will be used to store the dataset. You can use the default value.
+- `volume_size`: Size of the Kubernetes Persistent Volume Claims (PVC) that will be used to store the dataset. Normally, for 1K data, we need 3G volume.
```
adason: I will reword this to:

If you plan to use datasets from SynthDet(add url) project that include more than X images, you should increase this value accordingly. The project requires ~3GiB storage per 1K images.

BlairLee: Update the tutorial.

kubeflow/pipelines.py
```diff
@@ -387,13 +393,14 @@ def train_on_synthetic_and_real_dataset(
...
     config: str = "datasetinsights/configs/faster_rcnn_fine_tune.yaml",
+    override: str = "",
     checkpoint_file: str = (
         "https://storage.googleapis.com/datasetinsights/models/Synthetic"
         "/FasterRCNN.estimator"
     ),
     tb_log_dir: str = "gs://<bucket>/runs/yyyymmdd-hhmm",
     checkpoint_dir: str = "gs://<bucket>/checkpoints/yyyymmdd-hhmm",
-    volume_size: str = "100Gi",
+    volume_size: str = "50Gi",
```
86sanj: Why are you reducing memory size?

BlairLee: Revert to 100Gi. Don't need to change the volume size.

86sanj (approved):

PR #146, opened 2021-02-18
update dependencies to newer versions for sprawl compatibility

zook-unity:
# Peer Review Information

upgraded `google-cloud-storage` to >=1.31.1 and `numpy` to >=1.18.2

# Pull Request Check List

no new tests or documentation.


zook-unity:
one note: i have not done any additional testing for cloud training and storage, &c.

datasetinsights/evaluation_metrics/average_recall_2d.py
```diff
@@ -111,7 +111,7 @@ class MeanAverageRecallAverageOverIOU(EvaluationMetric):
     TYPE = "scalar"
 
     IOU_THRESHOULDS = np.linspace(
-        0.5, 0.95, np.round((0.95 - 0.5) / 0.05) + 1, endpoint=True
+        0.5, 0.95, int(np.round((0.95 - 0.5) / 0.05)) + 1, endpoint=True
```
adason, line 114: Is this the issue (feature) that was fixed (introduced) in numpy 1.18+?

pyproject.toml
```diff
@@ -26,10 +26,10 @@ include = [
 [tool.poetry.dependencies]
 python = "^3.7"
 cython = "^0.29.14"
-google-cloud-storage = ">=1.24.1 <=1.28.1"
+google-cloud-storage = ">=1.31.1"
```
adason: I am assuming this [issue](https://github.com/Unity-Technologies/datasetinsights/pull/82) was fixed in `google-cloud-storage 1.31+?

pyproject.toml
```diff
@@ -26,10 +26,10 @@ include = [
...
 jupyter = "^1.0.0"
 kornia = "^0.1.4"
-numpy = ">=1.17 < 1.18"
+numpy = ">=1.18.2"
```
adason: Let's use `^1.18.2` or `>=1.18.2 <2.0.0` (they are equivalent syntex). This prevents unexpected major version bump.

datasetinsights/evaluation_metrics/average_recall_2d.py
zook-unity: yup. linspace expects an int but round returns a numpy float64. i'm guessing that used to get converted to an int before, but it's probably not a good practice to rely on implicit type conversion either way.

pyproject.toml
zook-unity: i did not know about that issue. it looks like 1.31.1 merged those changes. we could do a later version if that matters, but i picked this to align with what sprawl uses currently

release notes for the fix: https://github.com/googleapis/python-storage/releases/tag/v1.31.1
`blob: base64 includes additional characters`

pyproject.toml
zook-unity: ah, good call! making that update.

pyproject.toml
adason: Good to know the issue is fixed. Let's also include `<2.0.0` in the requirements to make sure we don't accidentally bump the major version.

zook-unity: updated. anything look off?

adason (approved):

Issue #143, opened 2021-02-10
1.0 Release - IO and Statistics, remove experiments and model training


86sanj:
The current project supports features like ML experimentation, dataset IO, stats, and analysis on the notebook. In this release, we will remove all dependencies and code related model training and experiments to keep datasetinsights focused on io and statistics. This will reduce heavy decencies such as pytorch or tensorflow to make it easier for user to consume data from perception SDK without taking heavy dependencies. 

On high level, the 1.0 release of datasetinsights will contain the following modules

```
datasetinsights/
   io/
   stats/
   datasets/unity_perception/
notebooks/
```

## Checklist

- [x] Remove estimators, configs, trackers, kubflow pipeline, checkpoint
- [x] Prune dependencies 
- [x] Update README to reflect the current high level API. 
- [x] Update Perception_Statistics notebook, Remove yacs
- [x] Fix github action issue to prevent push 0.2.x tag to latest tag on dockerhub.

## How do I access model training code during this refactoring?

- Create a new branch after “Coco cat training and evaluation” commit, name 0.2.x.
- Create a tag on the master branch with the name 0.2.4. 
- Experimentation module: Users who want to use the train, evaluate can build the docker image on the code tagged with 0.2.4.
- Dataset IO, stats, and analysis on the notebook: Use the latest code from the master branch.

Any new changes or enhancements for the experimentation module will be done in the new repo but if any critical bug comes, checkout code from 0.2.x, fix the bug, and merge back to 0.2.x branch.
Release new tag from 0.2.x branch with the name 0.2.4.x

## What's the plan for future model training code? 

We will create a new repo that contain model training /experiment code. 
[Link to the new repo once it's ready]

## This work is split into multiple pull requests

- PR https://github.com/Unity-Technologies/datasetinsights/pull/142: Includes removal/fixes of the following modules which are related to model training and evaluation. 
Estimators
Train and Evaluate Command
Evaluation notebook
Configs yaml
Fixed docs rst and toctree dependency
IO module:
Tracker
Kfp output
Checkpoint
Config handler
Summary writer
Kubeflow pipelines.py
Evaluation metric
Corresponding unit test
docs

- PR https://github.com/Unity-Technologies/datasetinsights/pull/147: 
Remove tf and torch dependencies
Remove base and any public dataset
Remove any redundant library
Remove loader, draft notebook, protos
Updated readme
Updated docs

[...]

PR #155, opened 2021-03-04
Remove latest tag to stop override from 0.2.x

86sanj:
# Peer Review Information

- Removed latest tag to stop creating docker image with latest tag from 0.2.x branch.
Requirement: https://github.com/Unity-Technologies/datasetinsights/issues/143

# Pull Request Check List

### DIFF
--- a/.github/workflows/publish-docker-hub.yaml
+++ b/.github/workflows/publish-docker-hub.yaml
@@ -15,5 +15,4 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
           repository: unitytechnologies/datasetinsights
-          tags: latest
           tag_with_ref: true


.github/workflows/publish-docker-hub.yaml
```diff
@@ -15,5 +15,4 @@ jobs:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_ACCESS_TOKEN }}
           repository: unitytechnologies/datasetinsights
-          tags: latest
```
BlairLee, line 18: After this change, will we see the latest tag when we merge to the current master?

86sanj: This is merge request in 0.2.x branch. Master will have different code with latest tag.

adason: I am not sure if this will work. How do you test this out?

86sanj: In current implementation when this GitHub action runs, it publishes two docker images, one with latest tag (tags: latest), other with tag name( tag_with_ref: true).
- Keeping only tag_with_ref: true, will publish only one docker image with released tag. 
Reference: https://docs.github.com/en/actions/guides/publishing-docker-images

adason: How does this change in a branch fix the issue? Have you tested to make sure it will work as desired?

86sanj: - If we release from this branch it should publish only one docker with the release tag and this is what we want to achieve. With tag_with_ref: true,  only one docker image get published with release tag as per above documentation. 
- How to test this before merging the code:
1. One option is I can release from this PR branch with some test tag but it seems risky.
 2. Create a test project on docker hub and run this git hub action.
Please let me know which is preferred or any other method?

adason: Let's test it out using approach 2 to see if this change actually meet the desired behavior.

BlairLee (approved):

.github/workflows/publish-docker-hub.yaml
86sanj: Repo: https://github.com/sanjayuconn/ML_Deployment/tree/sanjayuconn-patch-1

- During the release(action trigger) the branch we select that's github yaml get executed.

adason: Can't see the repo. Can you make it open?

86sanj: https://github.com/sanjayuconn/literate-broccoli/tags

- Job from non master branch: https://github.com/sanjayuconn/literate-broccoli/runs/2082245697?check_suite_focus=true

- From master branch: https://github.com/sanjayuconn/literate-broccoli/runs/2082225878?check_suite_focus=true

- To confirm the execution please look at the content of run one-line script.

adason: excellent! This confirmed your hypothesis that a tag release on feature branch will trigger github action execution in a feature branch.

adason (approved):

PR #156, opened 2021-03-04
Add New metrics in Notebook

BlairLee:
# Peer Review Information

Add the following new metrics in `Perception_Statistics` Notebook:
- Foreground placement
- Foreground scale
- Foreground positions
- Lighting (color, pos, intensity, enabled states)
- Camera info (saturation, contrast, vignette amount, depth of field start/end/radius)

### DIFF
--- a/notebooks/Perception_Statistics.ipynb
+++ b/notebooks/Perception_Statistics.ipynb
@@ -111,13 +111,12 @@
    "outputs": [],
    "source": [
     "from datasetinsights.stats.statistics import RenderedObjectInfo\n",
-    "import datasetinsights.datasets.unity_perception.metrics as metrics\n",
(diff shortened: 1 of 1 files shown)


notebooks/Perception_Statistics.ipynb
```diff
@@ -230,6 +229,539 @@
     "    ))"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## SynthDet Statistics\n",
```
adason: Let's not add SynthDet specific statistics into the datasetinsights perception package. We should consider moving these statistics to the SynthDet project instead.

BlairLee: We could discuss which statistics should be in this package in the sync meeting.

BlairLee:
This PR should be submitted to SynthDet project.

PR #172, opened 2021-06-21
Orthographic camera

leopoldo-zugasti:
# Peer Review Information
By adding orthographic=True it computes the world point to pixel assuming the projection matrix is for orthographic projection
image

### DIFF
--- a/datasetinsights/datasets/synthetic.py
+++ b/datasetinsights/datasets/synthetic.py
@@ -27,17 +27,19 @@ def read_bounding_box_3d(annotation, label_mappings=None):
 
     for b in annotation:
         label_id = b["label_id"]
-        translation = b["translation"]
-        translation = [translation["x"], translation["y"], translation["z"]]
-        size = b["size"]
-        size = [size["x"], size["y"], size["z"]]
+        translation = (
+            b["translation"]["x"],
+            b["translation"]["y"],
+            b["translation"]["z"],
+        )
+        size = (b["size"]["x"], b["size"]["y"], b["size"]["z"])
         rotation = b["rotation"]
         rotation = Quaternion(
-            b=rotation["x"], c=rotation["y"], d=rotation["z"], a=rotation["w"]
+            x=rotation["x"], y=rotation["y"], z=rotation["z"], w=rotation["w"]
         )
 
-        if label_mappings and label_id not in label_mappings:
-            continue
+        # if label_mappings and label_id not in label_mappings:
+        #    continue
         box = BBox3D(
             translation=translation,
             size=size,
--- a/datasetinsights/datasets/synthetic.py
+++ b/datasetinsights/datasets/synthetic.py
@@ -38,8 +38,8 @@ def read_bounding_box_3d(annotation, label_mappings=None):
             x=rotation["x"], y=rotation["y"], z=rotation["z"], w=rotation["w"]
         )
 
-        # if label_mappings and label_id not in label_mappings:
-        #    continue
+        if label_mappings and label_id not in label_mappings:
+            continue
         box = BBox3D(
             translation=translation,
             size=size,
--- a/datasetinsights/stats/visualization/bbox3d_plot.py
+++ b/datasetinsights/stats/visualization/bbox3d_plot.py
@@ -129,7 +129,7 @@ def add_single_bbox3d_on_image(
 
 
 def _project_pt_to_pixel_location(pt, projection, img_height, img_width):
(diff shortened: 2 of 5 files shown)


datasetinsights/datasets/synthetic.py
```diff
@@ -27,17 +27,19 @@ def read_bounding_box_3d(annotation, label_mappings=None):
...
+        size = (b["size"]["x"], b["size"]["y"], b["size"]["z"])
         rotation = b["rotation"]
         rotation = Quaternion(
-            b=rotation["x"], c=rotation["y"], d=rotation["z"], a=rotation["w"]
+            x=rotation["x"], y=rotation["y"], z=rotation["z"], w=rotation["w"]
         )
 
-        if label_mappings and label_id not in label_mappings:
-            continue
+        # if label_mappings and label_id not in label_mappings:
```
BlairLee: If it's removed, it will get an error when there is something wrong with the label id. Or you can handle label_id out of defined labels case here.

datasetinsights/stats/visualization/plots.py
```diff
@@ -78,7 +78,7 @@ def _process_label(bbox, label_mappings=None):
         return label
 
 
-def plot_bboxes3d(image, bboxes, projection, colors=None):
+def plot_bboxes3d(image, bboxes, projection, colors=None, orthographic=False):
```
BlairLee, line 81: We don't have unittests for this method. Can you add it in `test_visual.py`?

datasetinsights/datasets/synthetic.py
StevenBorkman: I agree with @BlairLee on this, I wouldn't remove this error check unless you're 100% sure that the data is always going to be there. This check should not be hurting anything, this isn't performant code to start with

StevenBorkman (dismissed):
I agree with @BlairLee's comments, other than that, nothing else.

datasetinsights/datasets/synthetic.py
leopoldo-zugasti: Yeah I think I removed it to test something, there is no reason not to have it

datasetinsights/stats/visualization/plots.py
leopoldo-zugasti: Yeah I'll do that aswell

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -80,7 +80,7 @@ def _add_single_bbox3d_on_image(
 
 
 def add_single_bbox3d_on_image(
-    image, box, proj, color=None, box_line_width=2,
+    image, box, proj, color=None, box_line_width=2, orthographic=False,
```
adason, line 83: Update docstrings according to this change.

datasetinsights/stats/visualization/bbox3d_plot.py
```diff
@@ -161,3 +167,43 @@ def _project_pt_to_pixel_location(pt, projection, img_height, img_width):
             int((_pt[1] * img_height) / 2.0 + (img_height * 0.5)),
         ]
     )
+
+
+def _project_pt_to_pixel_location_orthographic(
```
adason, line 173: Add unittest for this method.

tests/test_bbox.py
```diff
@@ -59,3 +60,26 @@ def test_project_pt_to_pixel_location():
...
+    assert pixel_loc[1] == 240
+
+    # more interesting case
+    pt = [0.3, 0, 0]
+    proj = numpy.array([[0.08951352, 0, 0], [0, 0.2, 0], [0, 0, -0.0020006]])
+
+    pixel_loc = _project_pt_to_pixel_location_orthographic(
+        pt, proj, img_height, img_width
+    )
+    assert pixel_loc[0] == 328
```
BlairLee: How do you calculate the number 328? If we change the `img_height` and `img_width`, shall we also change the assertion?
You can add some simple calculations in the test code to automatically check it.

leopoldo-zugasti: Something like this?:
`assert pixel_loc[0] == int((proj[0][0] * pt[0] + 1) * 0.5 * img_width) # 328`
`assert pixel_loc[1] == img_width // 2`
@BlairLee

BlairLee: Exactly. If the assertion can be calculated easily, I recommend doing so.

BlairLee (approved):

adason (approved):

PR #180, opened 2021-09-20
COCO Pose and Stats

Saurav-D:
# Peer Review Information

- Added code for generating plots for coco pose and stats.

# Pull Request Check List


adason:
@Saurav-D From very high level, I think per visualization chart can be broken down into at least two steps: transformation and plots. You should be able to write tests against each step. Let's try to organize coco pose stats/visualization around this concept.

For example, if you are going to do COCO bounding box heat map plot, I would recommend trying to break it down into:

```
def transform(bboxes: List[BBox2D], width: float, height: float) -> pd.DataFrame:
  # transform a list of bounding boxes to suitable formats in pandas Dataframe (or any data container that is easier to plot)

def test_transform()
    input = [BBox2D(...), Bbox2D(...),...]
    width = 640
    height = 480
    expected = pd.DataFrame(...)

    assert transform(input, width, height) == expected

def heatmap_plot(data: pd.DataFrame):
   # in plot, don't do any data transformation here. The plot method should only care about visualization tweaks (axis, labels, legends,...) 
    sns.plot(...)


def test_heatmap_plot():
    # mock out sns.plot() method. make sure it's called with the expected plotting parameters. 

```

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,146 @@
...
+            relative_size = math.sqrt(area_size)
+            bbox_relative_size.append(relative_size)
+
+    for key in kpt_dict.keys():
+        kpt_dict[key] = kpt_dict[key] / total_instances
+
+    return kpt_dict, bbox_dict, bbox_heatmap, bbox_relative_size
+
+
+def convert_coco_obj_to_df(coco_obj, cat_ids, max_frames=None):
```
adason: There are a few issue about this method:
1. It's not clear what's the input/output here. First of all, I will not pass coco_obj here. Since you are primary dealing with annotations, I would extract the raw annotations out. A few good examples would be: list of bounding boxes, list of images where each image contain multiple bounding boxes, list of json records. You have to think about: how can I test this convert/transform method? 
2. This method has multiple outputs. Can you break it down to two separate method? Do they need to be coupled?

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+def _get_bbox_dict():
+    bbox_dict = {}
+    for key in COCO_AREA_DICT:
+        bbox_dict[key] = 0
+    return bbox_dict
+
+
+def get_stats(img_ids, cat_ids, coco_obj):
```
adason: This method return a tuple of 4 objects. Can you break it down into two methods? four methods?

datasetinsights/stats/visualization/coco_pose.py
```diff
@@ -0,0 +1,305 @@
...
+    "left_hip",
+    "right_hip",
+    "left_knee",
+    "right_knee",
+    "left_ankle",
+    "right_ankle",
+]
+
+
+def _load_annotations_from_json(json_file):
```
adason: Can the loading being handled in other modules other than visualization?

datasetinsights/stats/visualization/coco_pose.py
```diff
@@ -0,0 +1,305 @@
...
+        ax.patch.set_edgecolor("black")
+
+        ax.patch.set_linewidth("1")
+
+        # Invert axes
+        ax.invert_yaxis()
+
+        ax.set_aspect("equal")
+
+        if fig_dest:
```
adason: Let's decouple saving to file from the "generate plot" itself. The plot generate method should not have any side effects like this. Otherwise, it going to be very hard to write unittests.

datasetinsights/stats/visualization/coco_pose.py
```diff
@@ -0,0 +1,305 @@
...
+            plt.savefig(
+                os.path.join(fig_dest, f"{name}_heatmap.png"),
+                bbox_inches="tight",
+                pad_inches=0.15,
+            )
+
+        plt.show()
```
adason: Let's simply return a figure objet and have "plt.show" called somewhere else.

datasetinsights/stats/coco_stats.py
86sanj: - Prefer keyword arguments over positional.
- Pass the required data from coco object instead of passing whole object.
- Add doc string
- Decouple into at least 3 methods which can have responsibilities to handle Image, person and stats.

86sanj (commented):
Can we decouple into 3 modules:
1. Data transformation
2. Coco stats
3. Visualization: 
- Generate different plots, show plot and save plot
This way it would be easy to maintain, test and extend each of the modules.

datasetinsights/io/coco.py
```diff
@@ -0,0 +1,8 @@
+import json
+
+
+def load_coco_annotations(annotation_file):
+    f = open(annotation_file)
```
86sanj: 1. Use "with open"[(PEP 0343)](https://www.python.org/dev/peps/pep-0343/), as it close the open file automatically or close the file manually if you using only open.
2. What if file does not exist or not valid json? 
3. Docs string missing.

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+    2: "medium",
+    3: "large",
+    4: "96-128",
+    5: "128-256",
+    6: "256-512",
+    7: "512-inf",
+}
+
+
+def _get_empty_bbox_heatmap(coco_img_ids, coco_obj):
```
86sanj: 1. How coco_img_ids is different than coco_obj.getImgIds()? Can we simplify to have one parameter?

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+    # calculate bbox shape according to largest image in the dataset
+    max_height, max_width = 0, 0
+    for idx in coco_img_ids:
+        img = coco_obj.loadImgs(idx)[0]
```
86sanj: What if there is no data at 0th index? can we do data check before fetching?

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+        max_height = max(img["height"], max_height)
+        max_width = max(img["width"], max_width)
+
+    bbox_heatmap = np.zeros([max_height, max_width, 1])
+    return bbox_heatmap
+
+
+def _get_labeled_kpt_dict(coco_obj: pycocotools.coco.COCO):
+    kpt_dict = {}
+    coco_kpts = coco_obj.loadCats(coco_obj.getCatIds())[0]["keypoints"]
```
86sanj: What if there is no data at 0th index? can we do data check before fetching?

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+        bbox_dict[key] = 0
+    return bbox_dict
+
+
+def _get_bbox_relative_size(bbox_area, image_h, image_w):
+    bbox_relative_size = math.sqrt(bbox_area / (image_h * image_w))
+    return bbox_relative_size
+
+
+def get_labeled_keypoints(annotation_file):
```
86sanj: 1. Why we are coupling annotation loading with getting the key points? can we move to "io" directory? 
2. Whats the purpose of this method? Refer this for docstring([PEP-257](https://www.python.org/dev/peps/pep-0257/)).

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+    coco = COCO(annotation_file)
+    labeled_kpt_dict = _get_labeled_kpt_dict(coco_obj=coco)
+    keypoints = list(labeled_kpt_dict.keys())
+    img_ids = coco.getImgIds()
+    cat_ids = coco.getCatIds()
+    total_instances = 0
+
+    for idx in tqdm(img_ids):
+        img = coco.loadImgs(idx)[0]
```
86sanj: - [Same as above]: positional index without checking.

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,237 @@
...
+                if kp_visibility_flags[i] != 0:
+                    kp_visibility_flags[keypoints[i]] += 1
+
+    for key in labeled_kpt_dict.keys():
+        labeled_kpt_dict[key] = labeled_kpt_dict[key] / total_instances
+
+    return labeled_kpt_dict
+
+
+def get_bbox_heatmap(annotation_file):
```
86sanj: 1. [Same as above]: Coupling File loading and data processing.
2. Can we make sure that box computation is happening only for human? what if user loads file with many category?
3. Is it possible to implement in a way that box computation can work for any category? i.e flexible to extend for any category?
4.  Missing docstring.

tests/test_coco_stats.py
```diff
@@ -0,0 +1,153 @@
+from pathlib import Path
+from unittest.mock import MagicMock, Mock, patch
+
+import pytest
```
salehe-ee, line 4: Got the following error missing library
```
Traceback (most recent call last):
  File "tests/test_coco_stats.py", line 4, in <module>
    import pytest
ModuleNotFoundError: No module named 'pytest'
```

Saurav-D: Did you install all the requirements from `pyproject.toml`?

datasetinsights/io/coco.py
```diff
@@ -0,0 +1,25 @@
+import matplotlib.pyplot as plt
+from pycocotools.coco import COCO
+
+
+def load_coco_annotations(annotation_file: str) -> COCO:
```
adason, line 5: This one-line method is probably overkill, unless you have extra validation applied here. It's almost the same code `load_coco_annotations("file.json")` v.s. `COCO("file.json")`.

```
x
```

datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,131 @@
...
+
+from datasetinsights.io.coco import load_coco_annotations
+from datasetinsights.stats.coco_stats import (
+    load_annotations_for_single_img,
+    load_image_from_img_ann,
+    load_img_ann_for_single_image,
+)
+
+
+def display_single_img(
```
adason, line 14: I will recommend using this interface design:

```
def display_single_image(image, annotation, **display_params):
  ...
  # parse display parameters here (e.g. dpi, axis on/off, ..)
  fig, ax = plt.subplots(dpi=100)
  ax.axis("off")
  ax.imshow(image)
```

datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+    img_ann = load_img_ann_for_single_image(coco_obj=coco, img_id=img_id)
+    img = load_image_from_img_ann(img_annotation=img_ann, data_dir=data_dir)
+    fig, ax = plt.subplots(dpi=100)
+    ax.axis("off")
+    ax.imshow(img)
+    return fig
+
+
+def display_ann_for_single_img(
```
adason: Maybe you can combine this method with `display_single_image` and make `annotations` optional parameter, default to `None` (does not show annotations).

datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,131 @@
...
+    )
+    img = load_image_from_img_ann(img_annotation=img_ann, data_dir=data_dir)
+    fig, ax = plt.subplots(dpi=100)
+    ax.axis("off")
+    ax.imshow(img)
+    coco.showAnns(annotations, draw_bbox=True)
+    return fig
+
+
+def display_ann_for_all_img(
```
adason, line 56: Let's make this method 

```
def display_multiple_images(images: list, annotations: Optional[list] = None, nrows: int, ncols: int):
    # do some checks on the size of images and nrows
    fig, *axs = plt.subplots(nrows, ncols)
    for i, ax in enumerate(axs):
        ax.imshow(images[i])
        ....

```

datasetinsights/stats/coco_stats.py
```diff
@@ -0,0 +1,475 @@
...
+                if kp_visibility_flags[i] != 0:
+                    labeled_kpt_dict[keypoints[i]] += 1
+
+    for key in labeled_kpt_dict.keys():
+        labeled_kpt_dict[key] = labeled_kpt_dict[key] / total_instances
+
+    return labeled_kpt_dict
+
+
+def get_bbox_heatmap(
```
adason, line 225: Let's prevent passing an instance of `coco_obj` around all methods. I recommend using primative python data structures (list, dictionary, nested dictionary,...) or python `datacalss` to represent data loaded. This will allow you to extend dataset statistics beyond COCO format, without re-implement every single method for dataset in different format (open-close principle)

adason (changes requested):

adason:
My biggest concern is around the interface design here. Almost all modules in the PR use the `coco_obj` and pass around different methods to extract data, load image and visualize it. This violates open-close principle will make maintaining/extending this module hard. 

Can you come up with internal memory representation using python primitives (lists, dictionary, nested dictionary, or dataclass) and use this structure in all methods?

Saurav-D:
Closing this PR. Will create separate smaller PRs.

PR #178, opened 2021-09-03
Dataset Conversion Scripts - COCO 2D bounding boxes

adason:
# Peer Review Information

This change resolves #171 which convert 2D bounding boxes into COCO formats. 

# Pull Request Check List

- [x] Added **tests** for changed code.
- [x] Updated **documentation** for changed code.

PR #182, opened 2021-10-22
Added keypoints parsing support with the new parsing logic from `master`

salehe-ee:
# Peer Review Information

This change resolves remaining tasks from #178 for adding keypoint parsing support.
# Pull Request Check List


Saurav-D (commented):
Can we add test for COCO-Keypoints?

salehe-ee:
> add test

on it.

Saurav-D (approved):
LGTM!

datasetinsights/datasets/transformers/coco.py
```diff
@@ -167,3 +168,182 @@ def _categories(self):
...
+    def execute(self, output):
+        """Execute COCO Transformer
+        Args:
+            output (str): the output directory where converted dataset will
+              be stored.
+        """
+        self._copy_images(output)
+        self._process_instances(output)
+
+    def _copy_images(self, output):
```
adason, line 233: This violates DRY principle which is exactly the same copy of code from `COCOInstancesTransformer`. I would recommend using inheritance/composition pattern to prevent code duplication.

datasetinsights/datasets/transformers/coco.py
```diff
@@ -167,3 +168,182 @@ def _categories(self):
...
+                "file_name": f"camera_{capture_id}.png",
+                "height": height,
+                "width": width,
+                "id": capture_id,
+            }
+            images.append(record)
+
+        return images
+
+    def _annotations(self):
```
adason, line 277: The instance method `_annotations` and `_categories` might be the only difference in `COCOInstancesTransformer` and `COCOKeypointTransformer`. Can you use inheritance/composition to override the behavior for two different classes?

adason (changes requested):

adason (approved):
We agrees that the code refactor (using composition/inheritance) will be implemented later.

salehe-ee:
> We agrees that the code refactor (using composition/inheritance) will be implemented later.

Thank you. Are we good to merge these changes into `master`?

PR #184, opened 2021-11-01
Drop python 3.7 support & Remove tf-dataflow

adason:
# Peer Review Information

The package tornado is introduced as the dependencies, which seems to create circular dependencies when poetry tries to parse the dependency graph. This change seems to resolve the issue by dropping support for python 3.7. 

Screen Shot 2021-11-01 at 4 36 44 PM

Also, tf-dataflow is removed. This is not maintained and not in the critical path. 

# Pull Request Check List


Saurav-D (dismissed):

Saurav-D (commented):
I'm getting the following error when doing `poetry update`
Screen Shot 2021-11-09 at 10 25 41 PM

adason:
@Saurav-D What's the poetry version? Can you try `poetry update` in a virtual environment instead?

Saurav-D (approved):

PR #187, opened 2021-11-18
COCO Keypoints and BBox Viz

Saurav-D:
# Peer Review Information

Added COCO data visualization (KP and BBox). Also added notebook for it.

# Pull Request Check List


datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,78 @@
...
+from pycocotools.coco import COCO
+
+from datasetinsights.stats.coco_statistics import (
+    load_annotations_for_single_img,
+    load_image_from_img_ann,
+    load_img_ann_for_single_image,
+)
+
+
+def display_single_img(
```
adason, line 14: @Saurav-D Pleases decouple COCO api from this to allow other datasets in different format to reuse this method. I think the signature should be:

```
def display_single_image(image: str, annotation: Optional[list] = None):
    """
    Args:
        image (str): path the the image file
        annotation (list): list of 2d bounding boxes in this image
    """
    pass


def display_multiple_images(images: list, annotations: list, n_cols=3):
    """
    Args:
       images (list): list of path (string) to images files
       annotations (list): list of annotations for each image
    """
    pass

```

adason: Let's decouple data loading from COCO API from display images with bounding boxes so that it's open to extension for new dataset formats.

datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,78 @@
...
+        annotation_file (str): COCO annotation json file.
+        img_id (int): Image id of the image
+        data_dir (str): Directory where data(images) is located
+        show_annotation (bool): Show annotation for the image
+
+    Returns:
+        plt.Figure: Figure object
+
+    """
+    coco = COCO(annotation_file=annotation_file)
```
adason, line 32: This should be avoided in a function that's generic like this. Using COCO instances couple this method to datasets in COCO format.

datasetinsights/stats/visualization/coco_data_visualization.py
```diff
@@ -0,0 +1,78 @@
...
+    img_ann = load_img_ann_for_single_image(coco_obj=coco, img_id=img_id)
```
adason, line 33: This method should be called outside of `display_single_img`.

datasetinsights/stats/visualization/coco_data_visualization.py
Saurav-D: This file is specifically for coco-like datasets and visualization. Do we need to make it more generic for different datasets. Also the logic with loading and visualizing coco annotation might be different from other datasets.

PR #188, opened 2021-11-22
Changes to support legacy format updates

StevenBorkman:
# Peer Review Information

Slight changes to the perception format after SOLO work

# Pull Request Check List


JonathanHUnity:
I'm concerned about making this change with no backwards compatibility. PSP and the latest 0.9.0 releases of the Perception package will not be compatible with this new version of DatasetInsights. It looks like a rather small schema change - why can't we maintain the old format?

StevenBorkman:
> I'm concerned about making this change with no backwards compatibility. PSP and the latest 0.9.0 releases of the Perception package will not be compatible with this new version of DatasetInsights. It looks like a rather small schema change - why can't we maintain the old format?

I agree with you, so I reverted it back to the way it was before, although the way it was before did not follow any of the patterns already established for storing Vector3s or Quaternions

StevenBorkman:
I am actually going to cancel this pull request because the only changes now are just default values in user fillable fields

PR #200, opened 2022-03-14
Variance of Laplacian

Saurav-D:
# Peer Review Information

Added code for calculating:
- Laplacian of Image
- Variance of laplacian of image
- Var of lap of background on foreground based on the bounding box information.
- Var of lap of background on foreground based on the segmentation information.
- Added notebook for visualization

# Pull Request Check List

### DIFF
--- a/datasetinsights/stats/image_analysis/laplacian.py
+++ b/datasetinsights/stats/image_analysis/laplacian.py
@@ -20,38 +20,26 @@ def laplacian_img(img_path: str) -> np.ndarray:
     return laplacian
 
 
-def get_img_var_laplacian(img_path: str) -> np.ndarray:
-    """
-    Computes laplacian and returns the focus measure(ie variance for the image)
-    Args:
-        img_path (str): Path of image
-
-    Returns:
-        Variance of Laplacian of image
-    """
-    return laplacian_img(img_path).var()
-
-
 def get_bbox_var_laplacian(
-    laplacian: np.ndarray, x: float, y: float, w: float, h: float
+    laplacian: np.ndarray, x: int, y: int, w: int, h: int
 ) -> np.ndarray:
     """
     Calculates bbox's variance of Laplacian
     Args:
         laplacian (np.ndarray): Laplacian of the image
-        x (float): the upper-left coordinate of the bounding box
-        y (float): the upper-left coordinate of the bounding box
-        w (float): width of bbox
-        h (float): height of bbox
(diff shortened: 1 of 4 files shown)


datasetinsights/stats/image_analysis/laplacian.py
```diff
@@ -0,0 +1,85 @@
...
+        np.ndarray: numpy array of Laplacian of the image
+    """
+    image = cv2.imread(img_path)
+    gray = cv2.cvtColor(image, cv2.COLOR_BGR2GRAY)
+    laplacian = cv2.Laplacian(gray, cv2.CV_64F)
+    laplacian = laplacian.astype("float")
+    return laplacian
+
+
+def get_img_var_laplacian(img_path: str) -> np.ndarray:
```
davidwang-unity: this function may be an overkill here, compared `get_img_var_laplacian(img_path)` with `laplacian_img(img_path).var()`

datasetinsights/stats/image_analysis/laplacian.py
```diff
@@ -0,0 +1,85 @@
...
+        img_path (str): Path of image
+
+    Returns:
+        Variance of Laplacian of image
+    """
+    return laplacian_img(img_path).var()
+
+
+def get_bbox_var_laplacian(
+    laplacian: np.ndarray, x: float, y: float, w: float, h: float
```
davidwang-unity: why set `x y w h` as `float` here if we need to transfer them as `int` in the function?

datasetinsights/stats/image_analysis/laplacian.py
```diff
@@ -0,0 +1,85 @@
...
+        image
+        img_var_laplacian (np.ndarray): Variance of Laplacian of background
+        of the image
+
+    """
+    bbox_var_lap = []
+    img_laplacian = laplacian
+
+    for ann in annotations:
+        x, y, w, h = ann["bbox"]
```
davidwang-unity, line 64: maybe we can directly transfer `x y w h` as `int` type here?

datasetinsights/stats/image_analysis/laplacian.py
Saurav-D: Removed

datasetinsights/stats/image_analysis/laplacian.py
Saurav-D: Fixed

davidwang-unity (approved):
LGTM!

